PR #2, opened 2017-12-18
Basic tests

blegat:


blegat:
Coverage increased by 100% ahah :-D

schillic:
You set hard to beat standards - from now it gets only worse, we should stop here ;)

Issue #1, opened 2017-12-18
Constrained systems


blegat:
What do we do to encode constraints on state or input ?
For state constraints, we could do (or `DiscreteConstrainedLinearSystem` I don't know)
```julia
struct ConstrainedDiscreteLinearSystem{T, MT <: AbstractMatrix{T}, ST} <: AbstractDiscreteSystem
    A::MT
    X::ST
end
```
Then for the input we could have
```julia
struct DiscreteLinearControlSystem{T, MT <: AbstractMatrix{T}} <: AbstractDiscreteSystem
    A::MT
    B::MT
end
struct ConstrainedDiscreteLinearControlSystem{T, MT <: AbstractMatrix{T}, ST} <: AbstractDiscreteSystem
    A::MT
    X::ST
    B::MT
    U::ST
end
```
We could also have a type with unconstrained state and constrained input (or constrained state and unconstrained input) but it might be a bit too much, `ST` should have a value equivalent to the full unconstrained space to handle these cases.
We might consider having a different matrix type for `A` and `B` or a different  set type for `X` and `U`.
What do you think ?

schillic:
* I usually prefer `ConstrainedDiscreteLinearSystem` over `DiscreteConstrainedLinearSystem` because it specializes `DiscreteLinearSystem` (for the same reason I would actually prefer `LinearDiscreteSystem` over `DiscreteLinearSystem`).
* What is `ST` in a `DiscreteLinearControlSystem` - it is not used there?
* Agreed, having a type `Unconstrained` or such for `ST` sounds good (maybe even `Void`?). You typically do not create thousands of systems, so a little overhead is fine.
* With the above we could even drop `DiscreteLinearControlSystem` altogether because it is just a `ConstrainedDiscreteLinearControlSystem` where `ST == Unconstrained` (and then could again drop the `Constrained` prefix). But maybe this is going too far and we want to have more specific types...
* Different matrix types: bit of a mess, but it might be good to have the full generality here.

mforets:
let's defer "Different matrix types" as a feature that can be added later if u agree.

> What is ST in a DiscreteLinearControlSystem - it is not used there?

i think it was meant to be 

```
struct DiscreteLinearControlSystem{T, MT <: AbstractMatrix{T}, ST} <: AbstractDiscreteSystem
    A::MT
    B::MT
    U::ST
end
```
to model `x' = Ax + Bu`.

mforets:
[...]

PR #3, opened 2017-12-20
#1 - Constrained systems

mforets:


mforets:
note that `LazySet` is *not* required, but i found it convenient for writing tests. we can just update `.travis.yml` , "after success" to install `LazySet`.

schillic:
I am fine with using `LazySets` in the tests. It should be intuitive enough to read.

mforets:
the `osx` Travis build makes a big slow down (Running for 1 hr 36 min 38 sec), while it's < 3min for linux. then i've removed, please undo if u really want this or know why it's so slow -- see also [osx builds extremely slow to start](https://github.com/travis-ci/travis-ci/issues/7304)

src/continuous.jl
```diff
@@ -10,9 +10,83 @@ x' = A x.
...
+
+- `A` -- square matrix
+- `B` -- matrix
+"""
+struct LinearControlContinuousSystem{T, MT <: AbstractMatrix{T}} <: AbstractContinuousSystem
+    A::MT
+    B::MT
+end
+function LinearControlContinuousSystem{T, MT <: AbstractMatrix{T}}(A::MT, B::MT)
+    @assert size(A, 1) == size(B, 1)
```
blegat: We could check here that A is square by doing `@assert Base.LinAlg.checksquare(s.A) == size(B, 1)` instead

src/continuous.jl
```diff
@@ -10,9 +10,83 @@ x' = A x.
...
+- `U` -- input constraints
+"""
+struct ConstrainedLinearControlContinuousSystem{T, MT <: AbstractMatrix{T}, ST, UT} <: AbstractContinuousSystem
+    A::MT
+    B::MT
+    X::ST
+    U::UT
+end
+function ConstrainedLinearControlContinuousSystem{T, MT <: AbstractMatrix{T}, ST, UT}(A::MT, B::MT, X::ST, U::UT)
+    @assert size(A, 1) == size(B, 1)
```
blegat: same here

test/runtests.jl
```diff
@@ -1,4 +1,4 @@
-using Systems
+using Systems, LazySets
```
blegat: Why do we need LazySets ?

schillic: > note that LazySet is not required, but i found it convenient for writing tests. we can just update .travis.yml , "after success" to install LazySet.

blegat: Yes, I just noticed that comment ^ ^

blegat:
It seems to me that Continuous Control System is more common than Control Continuous System. I have done a quick google search. If I use either term, the first result is : https://link.springer.com/chapter/10.1007/978-1-84628-056-6_1 which uses "Continuous Control System".

schillic:
> It seems to me that Continuous Control System is more common than Control Continuous System.

@mforets probably chose this because of [my suggestion here](https://github.com/JuliaReach/Systems.jl/issues/1#issuecomment-352567896). I am of course fine with breaking this rule of thumb for existing terms.

mforets:
it's true that common usage in this context is "continuous-time systems" or "discrete-time systems", which are then specialized into more concrete classes. i'm favorable of accepting this slight abuse of notation and use "xyzContinuousSystem" and similarly for discrete, to have some homogeneity in the naming convention.

... we could go further to write "xyzCS" and "xyzDS" respectively but that's kind of discouraged in Julia.

on a separate note, thanks for the suggestions to improve the constructors, which i've incorporated. let me continue by adding the discrete analogues to this branch.

PR #4, opened 2017-12-27
Linear algebraic systems

mforets:
This PR adds yet another class from [HybridSystems.jl](https://github.com/blegat/HybridSystems.jl/blob/master/src/systems.jl#L71), the `LinearAlgebraic` ones, for the continuous and discrete cases.

### DIFF
--- a/src/continuous.jl
+++ b/src/continuous.jl
@@ -108,7 +108,7 @@ struct LinearAlgebraicContinuousSystem{T, MT <: AbstractMatrix{T}} <: AbstractCo
     E::MT
 end
 function LinearAlgebraicContinuousSystem{T, MT <: AbstractMatrix{T}}(A::MT, E::MT)
-    @assert Base.LinAlg.checksquare(A) && Base.LinAlg.checksquare(E)
+    @assert Base.LinAlg.checksquare(A) == Base.LinAlg.checksquare(E)
     return LinearAlgebraicContinuousSystem{T, MT}(A, E)
 end
 statedim(s::LinearAlgebraicContinuousSystem) = Base.LinAlg.checksquare(s.A)
@@ -134,7 +134,7 @@ struct ConstrainedLinearAlgebraicContinuousSystem{T, MT <: AbstractMatrix{T}, ST
     X::ST
 end
 function ConstrainedLinearAlgebraicContinuousSystem{T, MT <: AbstractMatrix{T}, ST}(A::MT, E::MT, X::ST)
-    @assert Base.LinAlg.checksquare(A) && Base.LinAlg.checksquare(E)
+    @assert Base.LinAlg.checksquare(A) == Base.LinAlg.checksquare(E)
     return ConstrainedLinearAlgebraicContinuousSystem{T, MT, ST}(A, E, X)
 end
 statedim(s::ConstrainedLinearAlgebraicContinuousSystem) = Base.LinAlg.checksquare(s.A)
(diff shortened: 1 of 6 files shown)


src/continuous.jl
```diff
@@ -90,3 +90,52 @@ end
...
+- `A` -- square matrix
+- `E` -- square matrix
+"""
+struct LinearAlgebraicContinuousSystem{T, MT <: AbstractMatrix{T}} <: AbstractContinuousSystem
+    A::MT
+    E::MT
+end
+function LinearAlgebraicContinuousSystem{T, MT <: AbstractMatrix{T}}(A::MT, E::MT)
+    @assert Base.LinAlg.checksquare(A) == Base.LinAlg.checksquare(E)
+    return LinearAlgebraicContinuousSystem{T, MT}(A, E)
```
schillic: Maybe define [such constructors inside the struct](https://docs.julialang.org/en/stable/manual/constructors/#Inner-Constructor-Methods-1).

src/continuous.jl
```diff
@@ -90,3 +90,52 @@ end
...
+end
+statedim(s::LinearAlgebraicContinuousSystem) = Base.LinAlg.checksquare(s.A)
```
schillic: No need to call the more expensive `Base.LinAlg.checksquare(s.A)` here: The constructor ensures that the matrix is square.

src/continuous.jl
```diff
@@ -90,3 +90,52 @@ end
...
+inputdim(s::LinearAlgebraicContinuousSystem) = 0
```
schillic: Minor: You can omit the name `s` here.

src/continuous.jl
mforets: if it's outside, this is a bit more flexible in the sense that the user can provide more outer constructors.. just a thought..

src/continuous.jl
mforets: done (the `statedim` changes for older types are better a separate PR)

src/continuous.jl
```diff
@@ -111,7 +111,7 @@ function LinearAlgebraicContinuousSystem{T, MT <: AbstractMatrix{T}}(A::MT, E::M
     @assert Base.LinAlg.checksquare(A) == Base.LinAlg.checksquare(E)
     return LinearAlgebraicContinuousSystem{T, MT}(A, E)
 end
-statedim(s::LinearAlgebraicContinuousSystem) = Base.LinAlg.checksquare(s.A)
+statedim(s::LinearAlgebraicContinuousSystem) = size(A, 1)
```
schillic: `A` -> `s.A` (also below)

src/continuous.jl
```diff
@@ -90,3 +90,52 @@ end
...
+
+Continuous-time linear algebraic system of the form
+```math
+E x' = A x.
+```
+
+### Fields
+
+- `A` -- square matrix
+- `E` -- square matrix
```
blegat: The matrices do not need to be square. But they should have the same dimension though.

src/continuous.jl
schillic: For me the idea is to perform this check *always*. Breaking type assumptions is usually not "flexible" 🤕 A user can still provide more outer constructors, they just have to call the inner one.

src/continuous.jl
mforets: you're right,  i'll use `size()` for the check

PR #7, opened 2017-12-29
Initial Value Problems

mforets:


src/ivp.jl
```diff
@@ -0,0 +1,34 @@
...
+end
+statedim(ivp::InitialValueProblem) = statedim(ivp.s)
+inputdim(ivp::InitialValueProblem) = inputdim(ivp.s)
+
+"""
+    IVP
+
+`IVP` is an alias for `InitialValueProblem`.
+"""
+const IVP = InitialValueProblem
```
blegat: It seems your text editor does not end its file with a `\n` :-P

mforets: but you're right, i updated recently Atom and had lost my configs.. fixed now.

thanks you both for the feedback 👍

PR #10, opened 2018-01-11
Add identity systems

blegat:
One system to rule them all

### DIFF
--- a/src/continuous.jl
+++ b/src/continuous.jl
@@ -1,4 +1,33 @@
 """
+    ContinuousIdentitySystem <: AbstractContinuousSystem
+
+Trivial identity continuous-time system of the form
+```math
+x' = 0.
+```
+"""
+struct ContinuousIdentitySystem <: AbstractContinuousSystem
+    statedim::Int
+end
+statedim(s::ContinuousIdentitySystem) = s.statedim
+inputdim(s::ContinuousIdentitySystem) = 0
+
+"""
+    ConstrainedContinuousIdentitySystem <: AbstractContinuousSystem
+
+Trivial identity continuous-time system of the form
+```math
+x' = 0.
+```
+"""
+struct ConstrainedContinuousIdentitySystem{ST} <: AbstractContinuousSystem
+    statedim::Int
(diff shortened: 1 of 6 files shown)


src/continuous.jl
```diff
@@ -1,3 +1,32 @@
...
+struct ContinuousIdentitySystem <: AbstractContinuousSystem
+    statedim::Int
+end
+statedim(s::ContinuousIdentitySystem) = s.statedim
+inputdim(s::ContinuousIdentitySystem) = 0
+
+"""
+    ConstrainedContinuousIdentitySystem <: AbstractContinuousSystem
+
+Trivial identity continuous-time system of the form
```
schillic: Maybe add "with state constraints" in line with `ConstrainedLinearContinuousSystem`?
(same for the discrete system)

test/continuous.jl
```diff
@@ -1,3 +1,20 @@
...
+        s = ContinuousIdentitySystem(sd)
+        @test statedim(s) == sd
+        @test inputdim(s) == 0
+    end
+end
+
+@testset "Continuous constrained identity system" begin
+    for sd in 1:3
+        X = Line([1., -1], 0.) # line x = y
+        s = ConstrainedContinuousIdentitySystem(sd, X)
```
schillic: Should the constraint not be of dimension `sd`? (Maybe I do not understand how this works.)
(same for the discrete system)

blegat: Indeed ^^

test/continuous.jl
```diff
@@ -7,12 +7,10 @@
 end
 
 @testset "Continuous constrained identity system" begin
-    for sd in 1:3
-        X = Line(ones(sd), 0.)
```
schillic: Sorry, I realized this only later: You can use `Singleton(ones(sd))` for this.

test/continuous.jl
mforets: yes, it was my fault -- i added a 2D `Line` but it was meant to be hyperplanes. anyways, `Singleton` is ok.

PR #16, opened 2018-03-10
Input iterators

### LINKED ISSUE #15: Input iterators
In applications the systems can receive inputs which are constant or which vary over time. This is to propose some functionality to make defining inputs of different nature easy and efficiently within `Systems.jl`. 

(Notice that i'm not proposing to "tie" existent systems definitions with new input types.)

```julia
abstract type AbstractInput end

Base.start(::AbstractInput) = 1

# input that remains constant in time
struct ConstantInput{UT} <: AbstractInput
    U::UT
end

Base.next(inputs::ConstantInput, state) = (inputs.U, state + 1)
Base.done(inputs::ConstantInput, state) = state > 1
Base.length(inputs::ConstantInput) = 1

# add this (?)
next_input(inputs::ConstantInput, state::Int) = inputs.U
next_input(inputs::ConstantInput) = inputs.U
Base.*(M::AbstractMatrix, input::ConstantInput) =  ConstantInput(M * input.U)

# input that may vary with time
struct TimeVaryingInput{UT} <: AbstractInput
    U::AbstractVector{<:UT}  # input sequence
end

Base.next(inputs::TimeVaryingInput, state) = (inputs.U[state], state + 1)
Base.done(inputs::TimeVaryingInput, state) = state > length(inputs.U)
Base.length(inputs::TimeVaryingInput) = length(inputs.U)

# add this (?)
next_input(inputs::TimeVaryingInput, state::Int) = inputs.U[state]
```

thoughts?

mforets:
Closes #15.

### DIFF
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -8,7 +8,10 @@ DocTestFilters = [r"[0-9\.]+ seconds \(.*\)"]
 
 ## Features
 
-Types for mathematical systems modeling.
+- Generic and flexible systems definitions, while being fast and type stable.
+- Types for mathematical systems modeling: continuous, discrete, controlled systems,
+linear algebraic, etc.
+- Iterator interfaces to handle constant or time-varying inputs. 
 
 ## Library Outline
 
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -9,9 +9,9 @@ DocTestFilters = [r"[0-9\.]+ seconds \(.*\)"]
 ## Features
 
 - Generic and flexible systems definitions, while being fast and type stable.
-- Types for mathematical systems modeling: continuous, discrete, controlled systems,
+- Types for mathematical systems modeling: continuous, discrete, controlled,
 linear algebraic, etc.
-- Iterator interfaces to handle constant or time-varying inputs. 
+- Iterator interfaces to handle constant or time-varying inputs.
 
 ## Library Outline
 
--- a/src/inputs.jl
+++ b/src/inputs.jl
@@ -47,6 +47,13 @@ Base.IsInfinite()
 julia> for ci in c print(2*ci) end
 -1//1
 ```
+
+The elements of this input are rational numbers:
+
+```jldoctest constant_input
+julia> eltype(c)
+Rational{Int64}
+```
 """
 struct ConstantInput{UT} <: AbstractInput
     U::UT
@@ -55,6 +62,7 @@ end
 Base.next(input::ConstantInput, state) = (input.U, state + 1)
 Base.done(input::ConstantInput, state) = state > 1
 Base.length(input::ConstantInput) = Base.IsInfinite()
+Base.eltype(::Type{ConstantInput{UT}}) where {UT} = UT
 
 """
     nextinput(input::ConstantInput, state::Int)
@@ -84,7 +92,7 @@ Type representing an input that may vary with time.
 ### Examples
 
 The varying input holds a vector of elements and its length equals the number
-of elements in the vector.
+of elements in the vector. Consider an input given by a vector of rational numbers:
 
 ```jldoctest varying_input
 julia> v = VaryingInput([-1//2, 1//2])
@@ -92,6 +100,9 @@ Systems.VaryingInput{Rational{Int64}}(Rational{Int64}[-1//2, 1//2])
 
 julia> length(v)
 2
+
+julia> eltype(v)
+Rational{Int64}
 ```
 
 The method `nextinput` receives a varying input and an index representing the state
@@ -127,6 +138,7 @@ end
 Base.next(input::VaryingInput, state) = (input.U[state], state + 1)
 Base.done(input::VaryingInput, state) = state > length(input.U)
 Base.length(input::VaryingInput) = length(input.U)
+Base.eltype(::Type{VaryingInput{UT}}) where {UT} = UT
 
 """
     nextinput(input::VaryingInput, state::Int)
--- a/src/inputs.jl
+++ b/src/inputs.jl
@@ -24,8 +24,7 @@ Type representing an input that remains constant in time.
 ### Examples
 
 The constant input holds a single element and its length is infinite.
-To access the field `U`, you can use the method `next_input`,
-and a `for` loop as well:
+To access the field `U`, you can use the method `next_input`, or a `for` loop:
 
 ```jldoctest constant_input
 julia> c = ConstantInput(-1//2)
--- a/src/inputs.jl
+++ b/src/inputs.jl
@@ -98,7 +98,7 @@ Type representing an input that may vary with time.
 
 ### Examples
 
-The varying input holds a vector of elements and its length equals the number
+The varying input holds a vector and its length equals the number
 of elements in the vector. Consider an input given by a vector of rational numbers:
 
 ```jldoctest varying_input
@@ -112,10 +112,24 @@ julia> eltype(v)
 Rational{Int64}
 ```
 
-The method `nextinput` receives a varying input and an integer `n` and returns
-the first `n` elements of this input:
+Base's `next` method receives the input and an integer state and returns the
+input element and the next iteration state:
 
 ```jldoctest varying_input
+julia> next(v, 1)
+(-1//2, 2)
+
+julia> next(v, 2)
+(1//2, 3)
+```
+
+The method `nextinput` receives a varying input and an integer `n` and returns
+an iterator over the first `n` elements of this input (where `n=1` by default):
+
+```jldoctest varying_input
+julia> typeof(nextinput(v))
+Base.Iterators.Take{Systems.VaryingInput{Rational{Int64}}}
+
 julia> collect(nextinput(v, 1))
 1-element Array{Rational{Int64},1}:
  -1//2
--- a/src/inputs.jl
+++ b/src/inputs.jl
@@ -41,9 +41,6 @@ julia> nextinput(c)
 julia> nextinput(c)
 -1//2
 
-julia> length(c)
-Base.IsInfinite()
-
 julia> for ci in c print(2*ci) end
 -1//1
 ```
@@ -60,8 +57,8 @@ struct ConstantInput{UT} <: AbstractInput
 end
 
 Base.next(input::ConstantInput, state) = (input.U, state + 1)
-Base.done(input::ConstantInput, state) = state > 1
-Base.length(input::ConstantInput) = Base.IsInfinite()
+Base.done(::ConstantInput, state) = state > 1
+Base.iteratorsize(::Type{ConstantInput}) = Base.IsInfinite()
 Base.eltype(::Type{ConstantInput{UT}}) where {UT} = UT
 
 """
@@ -121,7 +118,7 @@ You can collect the inputs in an array, or equivalently use list comprehension,
 
 ```jldoctest varying_input
 julia> collect(v)
-2-element Array{Any,1}:
+2-element Array{Rational{Int64},1}:
  -1//2
   1//2
 
--- a/src/inputs.jl
+++ b/src/inputs.jl
@@ -1,0 +1,48 @@
+"""
+    AbstractInput
+
+Abstract supertype for all input types.
+"""
+abstract type AbstractInput end
+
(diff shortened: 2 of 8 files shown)


mforets:
We have `statedim`, `stateset`, ... => for consistency i would rather use `nextinput` instead of `next_input`. This getter function just returns the first component of `Base.next`.

src/inputs.jl
```diff
@@ -0,0 +1,144 @@
...
+-1//1
+```
+"""
+struct ConstantInput{UT} <: AbstractInput
+    U::UT
+end
+
+Base.next(input::ConstantInput, state) = (input.U, state + 1)
+Base.done(input::ConstantInput, state) = state > 1
+Base.length(input::ConstantInput) = Base.IsInfinite()
```
schillic: `Base.done` and `Base.length` are a bit inconsistent. Do we want that?
I see that it is useful for a finite iteration.

src/inputs.jl
```diff
@@ -0,0 +1,144 @@
...
+
+"""
+    next_input(input::ConstantInput, state::Int)
+
+Returns the input.
+
+### Input
+
+- `input` -- a constant input
+- `state`  -- the state of the iterator
```
schillic: space :)

src/inputs.jl
```diff
@@ -0,0 +1,144 @@
...
+
+### Output
+
+The input field of `input`.
```
schillic: I find "field" too implementation specific. What about "set" (as above) or "representation"? (same below for `VaryingInput`)

src/inputs.jl
```diff
@@ -0,0 +1,144 @@
...
+
+### Input
+
+- `input` -- a constant input
+
+### Output
+
+The input field of `input`.
+"""
+next_input(input::ConstantInput) = input.U
```
schillic: Maybe drop this method by adding a default value in the two-argument method?

src/inputs.jl
```diff
@@ -0,0 +1,144 @@
...
+
+Type representing an input that may vary with time.
+
+### Fields
+
+- `U` -- vector of input sets
+
+### Examples
+
+The varying input holds a vector of element and its length equals the number
```
schillic: elements

src/inputs.jl
```diff
@@ -0,0 +1,144 @@
...
+Base.length(input::VaryingInput) = length(input.U)
+
+"""
+    next_input(input::VaryingInput, state::Int)
+
+Returns the input field given a state.
+
+### Input
+
+- `inpus` -- a varying input
```
schillic: input

schillic:
> i would rather use nextinput instead of next_input

What about just `input`? Since you pass the iterator state, the "next" is implicit.

src/inputs.jl
blegat: `IsInfinite` should be returned by `iteratorsize`, not `length`. `length` should not be implemented

mforets:
> What about just input? 

I prefer `nextinput` , this is so close to`Base.next`. On a related note in the next commit i'd like to add your usage notes from [here](https://github.com/JuliaReach/Reachability.jl/blob/master/src/Systems/Systems.jl#L43) and [here](https://github.com/JuliaReach/Reachability.jl/blob/master/src/Systems/Systems.jl#L112).

src/inputs.jl
schillic: Note that you need `Type{}` (I think): `Base.iteratorsize(::Type{ConstantInput}) = Base.IsInfinite()`

schillic:
What about implementing [`Base.eltype`](https://docs.julialang.org/en/stable/manual/interfaces/)?
```julia
Base.eltype(::Type{ConstantInput{UT}}) = UT
Base.eltype(::Type{VaryingInput{UT}}) = UT
```

src/inputs.jl
mforets: Ok, but i don't really understand its purpose then. Now `length` returns a `MethodError`:

```julia
julia> c = ConstantInput(-1//2)
julia> length(c)

ERROR: MethodError: no method matching length(::Systems.ConstantInput{LazySets.Hyperrectangle{Float64}})
Closest candidates are:
  length(::SimpleVector) at essentials.jl:256
  length(::Base.MethodList) at reflection.jl:558
  length(::MethodTable) at reflection.jl:634
  ...
```

and in consequence 

```julia
julia> [ci for ci in c]
ERROR: MethodError: no method matching length(::Systems.ConstantInput{Rational{Int64}})
Closest candidates are:

  length(::SimpleVector) at essentials.jl:256
  length(::Base.MethodList) at reflection.jl:558
  length(::MethodTable) at reflection.jl:634
  ...
Stacktrace:
 [1] _array_for at ./array.jl:462 [inlined]
 [2] collect(::Base.Generator{Systems.ConstantInput{Rational{Int64}},##1#2}) at ./array.jl:476

julia> collect(c)
ERROR: MethodError: no method matching length(::Systems.ConstantInput{Rational{Int64}})
Closest candidates are:
  length(::SimpleVector) at essentials.jl:256
  length(::Base.MethodList) at reflection.jl:558
  length(::MethodTable) at reflection.jl:634
  ...
Stacktrace:
 [1] _similar_for at ./array.jl:407 [inlined]
 [2] _collect at ./array.jl:437 [inlined]
 [3] collect(::Systems.ConstantInput{Rational{Int64}}) at ./array.jl:431
```

Maybe this is intended for `IsInfinite()`, i'm asking.

src/inputs.jl
```diff
@@ -128,17 +129,17 @@ Base.done(input::VaryingInput, state) = state > length(input.U)
...
-    next_input(input::VaryingInput, state::Int)
+    nextinput(input::VaryingInput, state::Int)
 
 Returns the input field given a state.
 
 ### Input
 
-- `inpus` -- a varying input
-- `state` -- the state of the iterator
+- `inputs` -- a varying input
```
schillic: If you rename it here, you also have to rename it below.

src/inputs.jl
mforets: The `ConstantInput` is somehow analogue to [Base.Iterators.repeated](https://docs.julialang.org/en/latest/base/iterators/#Base.Iterators.repeated). 

So i tried the following implementation:

```julia
abstract type AbstractInput end
    
struct ConstantInput{UT} <: AbstractInput
    U::UT
end

Base.eltype(::Type{ConstantInput{UT}}) where {UT} = UT

Base.start(::ConstantInput) = nothing
Base.next(input::ConstantInput, state) = (input.U, nothing)
Base.done(::ConstantInput, state) = false

Base.IteratorSize(::Type{<:ConstantInput}) = Base.IsInfinite()
Base.IteratorEltype(::Type{<:ConstantInput}) = Base.HasEltype()

# retrieve the first n inputs
nextinput(input::ConstantInput, n::Int=1) = Base.Iterators.take(input, n)
```

Let `c = ConstantInput(-1//2)`.  The following tests all pass:

```julia
using Base.Test

let i = 0
    for j = c
        @test -1//2 == -1//2
        i += 1
        i <= 10 || break
    end
    @test i == 11
end

@test eltype(c) == Rational{Int64}
@test Base.IteratorSize(typeof(c)) == Base.IsInfinite()
@test Base.IteratorEltype(typeof(c)) == Base.HasEltype()
```

But `nextinput` doesn't work as i expected:

```julia
julia> nextinput(c, 4)  # ok
Base.Iterators.Take{ConstantInput{Rational{Int64}}}(ConstantInput{Rational{Int64}}(-1//2), 4)

julia> collect(nextinput(c, 4))  # why?
MethodError: no method matching length(::ConstantInput{Rational{Int64}})
Closest candidates are:
  length(::SimpleVector) at essentials.jl:256
  length(::Base.MethodList) at reflection.jl:558
  length(::MethodTable) at reflection.jl:634
  ...

Stacktrace:
 [1] _min_length(::ConstantInput{Rational{Int64}}, ::UnitRange{Int64}, ::Base.HasLength, ::Base.HasLength) at ./iterators.jl:14
 [2] _similar_for at ./array.jl:407 [inlined]
 [3] _collect at ./array.jl:437 [inlined]
 [4] collect(::Base.Iterators.Take{ConstantInput{Rational{Int64}}}) at ./array.jl:431
 [5] include_string(::String, ::String) at ./loading.jl:522
```

schillic: Why `Base.Iterators.take(input, n)`? I thought you wanted `Base.Iterators.repeated(input, n)`.

mforets:
Ready from my side.

src/inputs.jl
schillic: I first made a mistake here (now fixed). I do not understand why you do not need `input.U`. The manual says "An iterator that generates the value x forever." Can you explain?

mforets: The code is now consistent with "a `ConstantInput` generates the value x forever" since

```julia
for ci in c
    print(ci)
end
```
is an infinite loop, while before it replied with the first entry of `c`.

schillic: But in the implementation x is `input`. For me x should be `input.U`. We want to generate `U` forever, not the wrapper.

src/inputs.jl
```diff
@@ -0,0 +1,185 @@
...
+Type representing an input that remains constant in time.
+
+### Fields
+
+- `U` -- input set
+
+### Examples
+
+The constant input holds a single element and its length is infinite.
+To access the field `U`, you can use Base's `next` given a state, or the metho
```
blegat: metho -> method

docs/src/index.md
```diff
@@ -8,7 +8,10 @@ DocTestFilters = [r"[0-9\.]+ seconds \(.*\)"]
 
 ## Features
 
-Types for mathematical systems modeling.
+- Generic and flexible systems definitions, while being fast and type stable.
+- Types for mathematical systems modeling: continuous, discrete, controlled systems,
```
schillic: Remove the second "systems"?

src/inputs.jl
```diff
@@ -0,0 +1,185 @@
...
+
+julia> next(c, 2)
+(-1//2, nothing)
+
+julia> collect(nextinput(c, 4))
+4-element Array{Systems.ConstantInput{Rational{Int64}},1}:
+ Systems.ConstantInput{Rational{Int64}}(-1//2)
+ Systems.ConstantInput{Rational{Int64}}(-1//2)
+ Systems.ConstantInput{Rational{Int64}}(-1//2)
+ Systems.ConstantInput{Rational{Int64}}(-1//2)
```
schillic: This is exactly what I meant about the `repeated`: The expected behavior is that you get `-1//2` four times.
Compare to the `VaryingInput` example.

src/inputs.jl
```diff
@@ -0,0 +1,185 @@
...
+end
+
+Base.eltype(::Type{ConstantInput{UT}}) where {UT} = UT
+
+Base.start(::ConstantInput) = nothing
+Base.next(input::ConstantInput, state) = (input.U, nothing)
+Base.done(::ConstantInput, state) = false
+
+Base.IteratorSize(::Type{<:ConstantInput}) = Base.IsInfinite()
+Base.IteratorEltype(::Type{<:ConstantInput}) = Base.HasEltype()
```
schillic: This is the default, so you could drop this line (but we might as well keep it for clarity and for the case that the default changes).
On a side note: The [Julia docs](https://docs.julialang.org/en/stable/manual/interfaces/) are outdated here: They do not use capital letters for `IteratorEltype`.

src/inputs.jl
```diff
@@ -0,0 +1,185 @@
...
+Returns the first `n` elements of this input.
+
+### Input
+
+- `input` -- a constant input
+- `n`     -- (optional, default: `1`) the number of desired elements
+
+### Output
+
+An iterator of type `Base.Iterators.Take` that represents the first `n` elements
```
schillic: outdated

src/inputs.jl
mforets: Yes, I took it from the code base of `repeated`, you can open the "latest docs" branch: [IteratorEltype](https://docs.julialang.org/en/latest/base/collections/#Base.IteratorEltype).
For `VaryingInput` I also kept the definitions although they are also defaults.

schillic:
Good job!

PR #30, opened 2018-08-13
Polynomial ODEs

### LINKED ISSUE #27: Polynomial ODEs

mforets:
Closes #27.

- [x] polynomial system, cont time
- [x] polynomial system, cont time tests
- [x] polynomial system, discr time
- [x] polynomial system, discr time tests

### DIFF
--- a/src/continuous.jl
+++ b/src/continuous.jl
@@ -173,3 +173,46 @@ end
 statedim(s::ConstrainedLinearAlgebraicContinuousSystem) = size(s.A, 1)
 stateset(s::ConstrainedLinearAlgebraicContinuousSystem) = s.X
 inputdim(s::ConstrainedLinearAlgebraicContinuousSystem) = 0
+
+"""
+    PolynomialContinuousSystem
+
+Continuous-time polynomial system of the form
+```math
+x' = p(x).
+```
+
+### Fields
+
+- `p`        -- polynomial
+- `statedim` -- number of state variables
+"""
+struct PolynomialContinuousSystem{T} <: AbstractContinuousSystem
+    p::T
+    statedim::Int
+end
+statedim(s::PolynomialContinuousSystem) = s.statedim
+inputdim(s::PolynomialContinuousSystem) = 0
+
+"""
+    ConstrainedPolynomialContinuousSystem
+
+Continuous-time polynomial system with a constraint set,
+```math
+x' = p(x), x(t) ∈ \\mathcal{X}
+```
+
+### Fields
+
+- `p`        -- polynomial
+- `X`        -- constraint set
+- `statedim` -- number of state variables
+"""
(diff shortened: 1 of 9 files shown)


schillic:
Is `T` a good name for the polynomial type?

mforets:
> Is T a good name for the polynomial type?

How about `PT` or `FT`?

schillic:
Something with `P` makes more sense to me.

blegat:
`PT` or `P` I'd say

mforets:
thanks for commenting, this is ready from my side

src/continuous.jl
```diff
@@ -173,3 +173,46 @@ end
...
+    p::PT
+    statedim::Int
+end
+statedim(s::PolynomialContinuousSystem) = s.statedim
+inputdim(s::PolynomialContinuousSystem) = 0
+
+"""
+    ConstrainedPolynomialContinuousSystem
+
+Continuous-time polynomial system with a constraint set,
```
schillic: I find "with state constraints" clearer (we used this description for similar system types).
(same comment for the discrete version)

src/continuous.jl
```diff
@@ -173,3 +173,46 @@ end
...
+- `statedim` -- number of state variables
+"""
+struct ConstrainedPolynomialContinuousSystem{PT, ST} <: AbstractContinuousSystem
+    p::PT
+    statedim::Int
+    X::ST
+end
+statedim(s::ConstrainedPolynomialContinuousSystem) = s.statedim
+stateset(s::ConstrainedPolynomialContinuousSystem) = s.X
+inputdim(s::ConstrainedPolynomialContinuousSystem) = 0
```
schillic: Not related, but should we define `inputdim(s::AbstractSystem) = 0` for the abstract supertype, so we do not have to make these trivial definitions for every subtype?

test/continuous.jl
```diff
@@ -78,13 +78,20 @@ end
     @test inputdim(s) == 0
 end
 
-@testset "Initial value problem for a continuous constrained linear algebraic system" begin
-    A = [1. 1; 1 -1]
-    E = [0. 1; 1 0]
```
schillic: Why did you remove these tests?

mforets: that was unintended 😭

src/continuous.jl
mforets: it is true that most of the time we have `inputdim` equal to `0`  when there is no input. but i kind of prefer to have to explicitly write the `statedim`, `stateset`, `inputdim` since this makes it clear that this is what you intend for each type...

regarding code duplication what bothers me a bit is Discrete which mostly copy pastes Continuous. and also how different Continuous types are built on top of each other by adding more fields. maybe [Mixers.jl](https://github.com/rafaqz/Mixers.jl) can help..

PR #29, opened 2018-08-13
Updates for 0.7

mforets:
Closes #28.

- [x] use inner constructors if necessary
- [x] fix some doctests with `Matrix(::Transpose)`, that were failing because `A::MT` and `B::MT` fields are of the same type
-  [x] update travis script and add Compat to REQUIRE
- [x] new iteration protocol for `inputs.jl`
- [x] fix 0.6 (default constructors)

PR #31, opened 2018-08-14
#26 - Update installation instructions

### LINKED ISSUE #26: Update installation instructions
Now that this package is on METADATA, we should update the installation instructions.

mforets:
Closes #26.


schillic:
I would add this with #29.

PR #35, opened 2018-10-28
#25 - Affine systems with a control

### LINKED ISSUE #25: Affine systems
... of the form `x' = Ax + Bu + C` with constraints in the state space and the input set, and `C` a constant `n x 1` vector , would be useful for some instances. It can be called `ConstrainedAffineControlContinuousSystem`.

---

EDIT: 

- #32 added affine continuous `x' = Ax + b`
- #34 adds constrained affine `x' = A x + b, x(t) ∈ X`. 

This issue is to add `x' = Ax + Bu + c`, `x(t) ∈ X`, `c` a vector, `u(t) ∈ U`.

mforets:
Closes #25.


src/continuous.jl
```diff
@@ -148,6 +148,38 @@ statedim(s::ConstrainedAffineContinuousSystem) = length(s.b)
...
+- `U` -- input constraints
+"""
+struct ConstrainedAffineControlContinuousSystem{T, MT <: AbstractMatrix{T}, VT <: AbstractVector{T}, ST, UT} <: AbstractContinuousSystem
+    A::MT
+    B::MT
+    c::VT
+    X::ST
+    U::UT
+    function ConstrainedAffineControlContinuousSystem(A::MT, B::MT, c::VT, X::ST, U::UT) where {T, MT <: AbstractMatrix{T}, VT <: AbstractVector{T}, ST, UT}
+        @assert checksquare(A) == length(c)
```
blegat: should also be equal to `size(B, 1)`

PR #36, opened 2018-10-28
Fix size of U in tests

mforets:
Fixes an inconsistency (that i previously introduced 😲) for the size of inputs in the tests: the linear map (lazy or not) `B * U` should be commensurate. Since `size(B, 2) = 1` in the tests, one can use intervals or hyperrectangles in a single dimension interchangeably.


mforets:
i left the others, to have some mix, it doesn't really matter which set type is used in these particular tests

schillic:
¯\\\_(ツ)\_/¯

PR #37, opened 2018-10-28
#18 - Systems with outputs

### LINKED ISSUE #18: LTI systems type
In control books, (continuous) LTI systems are those of the form: `x' = Ax + Bu; y = Cx + Du`.

Proposal: define a `LinearTimeInvariantContinuousSystem` with fields `A, B, C, D` and another one `ConstrainedLinearTimeInvariantContinuousSystem` with fields `A, B, C, D, X, U`.

mforets:
Closes #18.

### DIFF
--- a/src/MathematicalSystems.jl
+++ b/src/MathematicalSystems.jl
@@ -67,4 +67,9 @@ export AbstractInput, ConstantInput, VaryingInput, nextinput
 
 include("inputs.jl")
 
+#==================================
+Systems with outputs
+===================================#
+include("maps.jl")
+
 end # module
--- a/src/MathematicalSystems.jl
+++ b/src/MathematicalSystems.jl
@@ -63,13 +63,28 @@ include("ivp.jl")
 #=====================
 Input related methods
 =====================#
-export AbstractInput, ConstantInput, VaryingInput, nextinput
+export AbstractInput,
+       ConstantInput,
+       VaryingInput,
+       nextinput
 
 include("inputs.jl")
 
 #==================================
-Systems with outputs
+Maps
 ===================================#
+export LinearMap,
+       AffineMap,
+       LinearControlMap,
+       AffineControlMap
+
+include("maps.jl")
+
+#=========================
+Systems with outputs
+==========================#
+export SystemWithOutput
+
 include("maps.jl")
 
 end # module
--- a/src/MathematicalSystems.jl
+++ b/src/MathematicalSystems.jl
@@ -14,8 +14,7 @@ export AbstractSystem,
        statedim,
        stateset,
        inputdim,
-       inputset,
-       outputdim
+       inputset
 
 include("abstract.jl")
 
@@ -75,6 +74,7 @@ include("inputs.jl")
 Maps
 ===================================#
 export outputmap,
+       outputdim,
        LinearMap,
        AffineMap,
        LinearControlMap,
--- a/src/MathematicalSystems.jl
+++ b/src/MathematicalSystems.jl
@@ -78,7 +78,9 @@ export outputmap,
        LinearMap,
        AffineMap,
        LinearControlMap,
-       AffineControlMap
+       ConstrainedLinearControlMap,
+       AffineControlMap,
+       ConstrainedAffineControlMap
 
 include("maps.jl")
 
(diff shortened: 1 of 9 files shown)


mforets:
i've hesitated about `...maps` because one can argue that `...DiscreteSystem` can be used for that purpose. perhaps we can make the `MT` in the output be optional, eg.

```julia
struct SystemWithOutput{ST<:AbstractSystem, MT}   # no restriction on MT
    system::ST
    outputmap::MT
end
```

src/maps.jl
```diff
@@ -0,0 +1,91 @@
...
+- `U` -- input constraints
+"""
+struct LinearControlMap{MT<:AbstractMatrix, UT} <: AbstractMap
+    A::MT
+    B::MT
+    U::UT
+end
+
+"""
+    AffineMap
```
schillic:
```suggestion
    AffineControlMap
```

schillic:
I do not understand what you want to say, sorry.

src/maps.jl
```diff
@@ -0,0 +1,91 @@
...
+
+An affine map
```
blegat: affine control

src/maps.jl
```diff
@@ -0,0 +1,91 @@
...
+"""
+struct AffineControlMap{MT<:AbstractMatrix, VT<:AbstractVector, UT} <: AbstractMap
+    A::MT
+    B::MT
+    c::VT
+    U::UT
+end
+
+"""
+    SystemWithOutput
```
blegat: Maybe add parameters here and add in the `Fields` section "of type `ST`"

src/maps.jl
```diff
@@ -0,0 +1,91 @@
...
+"""
+struct AffineMap{MT<:AbstractMatrix, VT<:AbstractVector} <: AbstractMap
+    A::MT
+    b::VT
+end
+
+"""
+    LinearControlMap
+
+An affine map
```
blegat: affine -> linear control

schillic:
>```julia
>struct LinearMap{T, MT<:AbstractMatrix} <: AbstractMap
>    A::MT
>end
>```

You need to use `T` here :wink:

blegat:
Yes it should be `AbstractMatrix{T}`

mforets:
i've added 2 more types `ConstrainedLinearControlMap` and `ConstrainedAffineControlMap` making `LinearControlMap` and `AffineControlMap` to be unconstrained (as it is the case for systems).

then i've added unit tests and docs.

new methods `outputdim` and `outputmap` have been defined for a `SystemWithOutput`. 

an LTI system can be constructed as a `LTISystem(A, B, C, D)` or a `LTISystem(A, B, C, D, X, U)` if one wants to specify both state and input constraints.

ready from my side.

src/MathematicalSystems.jl
```diff
@@ -14,7 +14,8 @@ export AbstractSystem,
        statedim,
        stateset,
        inputdim,
-       inputset
+       inputset,
+       outputdim
```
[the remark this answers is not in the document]
mforets: ok

src/outputs.jl
```diff
@@ -0,0 +1,87 @@
...
+- `D` -- matrix
+
+### Output
+
+A system with output such that the system is a linear control
+continuous system and the output map is a linear control map.
+"""
+function LinearTimeInvariantSystem(A, B, C, D)
+    system = LinearControlContinuousSystem(A, B)
+    outputmap = LinearControlMap(A, B)
```
blegat: It should be `C`, `D` here

mforets: oops

blegat: We should add a test to catch that

test/outputs.jl
```diff
@@ -2,21 +2,25 @@
...
     s = LinearTimeInvariantSystem(A, B, C, D)
     @test statedim(s) == 2
     @test inputdim(s) == 1
     @test outputmap(s) isa LinearControlMap
 
     # check alias
     s = LTISystem(A, B, C, D)
+
+    # check the LTI system's outputmap fields 
+    @test outputmap(s).A == C
```
schillic: You could add the analogous tests for `s.system`.

PR #55, opened 2019-02-01
#53 - Add ResetMap

### LINKED ISSUE #53: Add ResetMap
A `ResetMap <: AbstractMap` understood as the function that resets some components of the variable `x` to a prescribed value.

For example, `ResetMap(10, 9 => 0.)` that takes `x = (x1, ..., x9, x10)` and maps it to `(x1, ..., 0., x10)`.

mforets:
Closes #53.

### DIFF
--- a/test/maps.jl
+++ b/test/maps.jl
@@ -57,7 +57,7 @@ end
 end
 
 @testset "Reset map" begin
-    m = ResetMap(10, 9 => 0.)
+    m = ResetMap(10, Dict(9 => 0.))
     @test outputdim(m) == 10
 
     m = ResetMap(10, 2 => -1., 5 => 1.)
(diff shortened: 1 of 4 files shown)


test/maps.jl
```diff
@@ -55,3 +55,11 @@ end
     m = ConstrainedAffineControlMap(A, B, c, U)
     @test outputdim(m) == 2
 end
+
+@testset "Reset map" begin
+    m = ResetMap(10, 9 => 0.)
```
schillic: Maybe turn this test into a real `Dict{Int, Float64}`. The other test already covers the the automatic wrapping.
```julia
julia> typeof(9 => 0.)
Pair{Int64,Float64}
```

PR #57, opened 2019-02-01
#56 - Add islinear()

### LINKED ISSUE #56: Add apply method for maps
The idea is to add a function `apply(::AbstractMap, ::Any)` that implements the rule specified by the map. For example, `apply(m::LinearMap, x) = m.A * x`. Note that `x` is of type any: the product will dispatch on the types of A and x, to make it as generic as possible.

mforets:
Closes #56.

### DIFF
--- a/src/maps.jl
+++ b/src/maps.jl
@@ -14,6 +14,7 @@ struct IdentityMap <: AbstractMap
     dim::Int
 end
 outputdim(m::IdentityMap) = m.dim
+islinear(::IdentityMap) = true
 
 """
     LinearMap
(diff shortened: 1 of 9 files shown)


mforets:
Ready from my side.

src/maps.jl
```diff
@@ -182,3 +189,11 @@ end
 outputdim(m::ResetMap) = m.dim
 
 ResetMap(dim::Int, args::Pair{Int, <:N}...) where {N} = ResetMap(dim, Dict{Int, N}(args))
+
+function apply(m::ResetMap, x)
+    y = copy(x)
+    for index in keys(m.dict)
+        y[index] = m.dict[index]
```
schillic: This is the beginner's way to walk over a dictionary.
Never walk through the keys if you also access the values.
Use [`pairs`](https://docs.julialang.org/en/v1/base/collections/#Base.pairs) instead.

PR #58, opened 2019-02-01
#51 - Add islinear trait

### LINKED ISSUE #51: Add islinear trait
To dispatch reachability algorithms on systems we use an or clause, `if S <: LinearContinuousSystem || S <: LinearControlContinuousSystem || . . .` (see [default_operator](https://github.com/JuliaReach/Reachability.jl/blob/master/src/solve.jl#L4)).

Perhaps we can add in this library a function `islinear(S::AbstractSystem)` that returns `true` for all those cases where "Linear" is part of the type's name.

mforets:
Closes #51.

### DIFF
--- a/docs/src/lib/methods.md
+++ b/docs/src/lib/methods.md
@@ -35,3 +35,9 @@ nextinput
 outputdim
 outputmap
 ```
+
+## Traits
+
+```@docs
+islinear
+```
--- a/docs/src/lib/methods.md
+++ b/docs/src/lib/methods.md
@@ -39,5 +39,6 @@ outputmap
 ## Traits
 
 ```@docs
-islinear
+islinear(::AbstractSystem)
+islinear(::AbstractMap)
 ```
--- a/src/abstract.jl
+++ b/src/abstract.jl
@@ -48,6 +48,14 @@ Abstract supertype for all continuous system types.
 abstract type AbstractContinuousSystem <: AbstractSystem end
 
 """
+    islinear(s::AbstractSystem)
+
+Specifies if the rhs of system `s` are linear or not; this class also includes
+affine systems.
+"""
+function islinear(::AbstractSystem) end
+
+"""
     AbstractMap
 
 Abstract supertype for all map types.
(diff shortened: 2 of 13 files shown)


docs/src/lib/methods.md
```diff
@@ -34,4 +34,10 @@ nextinput
 ```@docs
 outputdim
 outputmap
+```
+
+## Traits
+
+```@docs
+islinear
```
schillic: Is this interesting? It will just list *all* methods here. I find this super confusing.

src/abstract.jl
```diff
@@ -47,6 +47,14 @@ Abstract supertype for all continuous system types.
 """
 abstract type AbstractContinuousSystem <: AbstractSystem end
 
+"""
+    islinear(s::AbstractSystem)
+
+Specifies if the rhs of system `s` are linear or not; this class also includes
+affine systems.
```
schillic: I do not understand this description.

blegat: rhs is ambiguous indeed, I would rather say whether the dynamic is described by linear equations.
Is an algebraic system linear ?

mforets:
> My approval is up to that I dislike islinear(::AffineMap) == true etc.

I think this is one case of a little abuse of notation that doesn't hurt :)

schillic:
> I think this is one case of a little abuse of notation that doesn't hurt :)

I do not see a problem in calling the function `isaffine`, but well...

mforets:
@schillic thanks for the prompt feedback. I'll wait if @blegat has further comments before merging.

src/abstract.jl
```diff
@@ -50,8 +50,23 @@ abstract type AbstractContinuousSystem <: AbstractSystem end
...
+
+We adopt the notion from [Section 2.7, 1]. For example, the system with inputs
+``x' = f(t, x, u) = A x + B u`` is linear, since the function ``f(t, ⋅, ⋅)`` is
+linear in ``(x, u)`` for each ``t ∈ \\mathbb{R}``. On the other hand,
+``x' = f(t, x, u) = A x + B u + c`` is affine but not linear, since it is not
+linear in ``(x, u)``.
+
+This function uses the information of the type, not the value. So, if a system
+type allows an instance that is not linear, it returns `false` by default.
+For example, polynomial systems can be linear not, hence `islinear`
```
blegat: "linear not" -> "nonlinear"

schillic: This already got fixed.

Issue #60, opened 2019-02-03
Maps with constraints


mforets:
The `src/maps.jl` defines a bunch of structures that subtype `AbstractMap`. As you can see in the table below, we are missing several combinations.

| Type | Constrains state | Constrains inputs | Is defined? |
|-----|----------------|------------------|------------|
|  IdentityMap   |                           |                                     |           x D|
|  ConstrainedIdentityMap   |            x               |                                     |          D |
|  LinearMap   |                           |                                     |           x D|
|  AffineMap   |                           |                                     |           x D|
|  LinearControlMap   |                           |                                     |           x D|
|  ConstrainedLinearMap   |          x                 |                                     |    D       |
|  ConstrainedAffineMap   |          x                 |                                     |      D     |
|  AffineControlMap   |                          |                                    |          x  D|
|  ConstrainedAffineControlMap   |          x                 |         x                            |     x (needs update)       |
|  ConstrainedLinearControlMap   |          x                 |         x                            |    x (needs update)      |

---

Here are two approaches to complete the table:

1. Add all the missing combinations one by one, e.g. `ConstrainedLinearMap` for `x -> Ax` with constraints on `x`, etc.

2. Add a new struct by composition,  defining maps with constraints, like:

```julia
struct ConstrainedMap{C, M<:AbstractMap}
    constraint::C
    map::M
end
```

For `...ContinuousSytem` and `...DiscreteSystem` we've been implementing approach 1.

I like approach 2 because it is more general. At the same time, it has less structure, because it does not speak about "constraint on what" (state or input).

schillic:
I prefer 2.

schillic:
Add `ConstrainedControlMap` which can also constrain the input.

schillic:
I think this has been resolved in #66.

PR #66, opened 2019-02-07
Add more maps 

### LINKED ISSUE #64: Add inputdim to maps

mforets:
Closes #64.

This PR adds the missing combinations of specialized map types, see the table in #60.

I also changed the behavior of some of the existing ones to match exactly the names of the corresponding systems. For example, in `master` we have that `ConstrainedAffineControlMap` has input constraints but no state constraints, althugh `ConstrainedAffineControlContinuousSystem` has both input and state constraints; then i let `ConstrainedAffineControlMap` to also have state constraints in this branch.

There are also minor docstrings tweaks to make them uniform throughout.

### DIFF
--- a/src/maps.jl
+++ b/src/maps.jl
@@ -1,9 +1,10 @@
 """
     IdentityMap
 
-An identity map
+An identity map,
+
 ```math
-x ↦ x
+    x ↦ x.
 ```
 
 ### Fields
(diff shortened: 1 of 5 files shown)


src/maps.jl
```diff
@@ -1,9 +1,10 @@
 """
     IdentityMap
 
-An identity map
+An identity map,
+
 ```math
-x ↦ x
+    x ↦ x.
```
blegat: Why is the indent needed ? Isn't it ignore in math mode ?

mforets: It is completely useless. The code looked nicer if i indented them  ¯\_(ツ)_/¯

mforets:
*Petit à petit, l’oiseau fait son nid.*

PR #70, opened 2019-02-09
Add map macro

mforets:
This is a proof of principle implementation of a `@map` macro using [Espresso.jl](https://github.com/dfdx/Espresso.jl) to make the matching. For the expected syntax, i used `->`:

```julia
julia> @map x -> [1 0; 0 0]*x
LinearMap{Int64,Array{Int64,2}}([1 0; 0 0])

julia> @map z -> [1 0; 0 0]*z
LinearMap{Int64,Array{Int64,2}}([1 0; 0 0])

julia> @map x -> [1 0; 0 0]*z
ERROR: ArgumentError: unable to match the given expression to a known map type
Stacktrace:
 [1] top-level scope at /Users/forets/.julia/dev/MathematicalSystems/src/macros.jl:49

julia> @map x -> [1 0; 0 0]*x + [2, 0]
AffineMap{Int64,Array{Int64,2},Array{Int64,1}}([1 0; 0 0], [2, 0])
``` 

Other things remain open, such as the syntax to define constraints, and the analogue macro for instances of ODE types.


blegat:
> I would prefer the following syntax:
> ```julia
> x' = f(x)
> ```

This is weird for a map, suppose `A` is not square, the output cannot be called `x'`

schillic:
> suppose `A` is not square, the output cannot be called `x'`

True, I have not thought about that case. Then I take it back.

src/macros.jl
```diff
@@ -0,0 +1,51 @@
+import Espresso.matchex
```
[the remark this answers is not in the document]
mforets: Thanks, i've updated the branch. I also had to update the lower level of Julia from `v0.7` to `v1.0` if we want to use the current version of Espresso.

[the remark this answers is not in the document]
mforets: > Use `using` instead of `import` if you do not want to extend the method.

The reason i used `import` instead of `using` is that i was only interested in one specific method, since `using` brings in all names exported by the module. I think either way is ok here.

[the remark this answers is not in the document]
schillic: > using brings in all names exported by the module

I meant `using Espresso.matchex` :wink:

mforets:
About point 2: certainly, it would be nice that `map` does not need to have the apply rules hard coded inside. If you let each map to register it's right hand side,

```julia
rhs(::Type{AffineMap}) = Meta.parse("_A * $x + _b")
```
trying to use it as the matching expression doesn't work because `$x` should be evaluated inside the scope of the macro, not earlier. So we can address this generalization in a follow up PR, if you agree.

To specify domains, i've thought about the syntax:

```julia
@map x -> A*x, x ∈ X
# or
@map(x -> A*x, x ∈ X)
```
but i don't know if one can pass `x ∈ X` to the macro as an expression (not as a bool)..

src/macros.jl
```diff
@@ -0,0 +1,51 @@
+using Espresso
```
mforets:
```suggestion
using Espresso: matchex
```

mforets:
Thanks. Let me add those easy ones before merging. How about the identity map?

```julia
@map x -> x   # what is the dimension?
```
in Julia v1.2 i think that we can do:

```julia
@map x -> I(2) * x
```
so we can just wait to add this until it's available.

schillic:
> what is the dimension?

Hm, tricky. The `IdentityMultple` approach is probably the best we have.
Some other ideas, but I am not convinced by either one:
```julia
x -> id(n)
x -> eye(n)
x -> x, dim = n
x -> x, n = n0
```
Then again, this map type is easy to write manually, so maybe we do not need to support it at all.

mforets:
Good. I'll try if i can get `@map x -> I(2) * x` to work, so we make use of the `IdentityMultiple`. If you like that we have an alias, such as `x -> x, dim = n`, i think it's a good idea and we can add it in a separate PR.

src/MathematicalSystems.jl
```diff
@@ -10,6 +10,8 @@ Identity operator
 =======================#
 include("identity.jl")
 
+export I
```
schillic: Is that the correct way? I thought you should do `import LinearAlgebra.I`.

src/macros.jl
```diff
@@ -1,13 +1,14 @@
...
-    map(ex)
+    map(ex, dom)
 
 Returns an instance of the map type corresponding to the given expression.
 
 ### Input
 
-- `ex` -- an expression defining the map, in the form of an anonymous function
+- `ex`  -- an expression defining the map, in the form of an anonymous function
+- `dom` -- additional optional arguments
```
schillic: What about a more generic name?

src/macros.jl
```diff
@@ -27,13 +28,52 @@ We can create an affine system as well:
...
+Additional arguments can be passed to `@map` using the function-call form, i.e.
+separating the arguments by commas, and using parentheses around the macro call.
+For exmple, an identity map of dimension 5 can be defined as:
+
+```jldoctest
+julia> @map(x -> x, dim=5)
+IdentityMap(5)
+```
+
+An identity map can alternatively been used by giving a the size of the identity
```
schillic: `been used by giving a` → `be created by giving`

src/MathematicalSystems.jl
mforets: since i use `LinearAlgebra.I` in the [new function](https://github.com/JuliaReach/MathematicalSystems.jl/pull/70/files#diff-4fe9194b62a81db935e115f91f14f135R69), that's fine i think.

schillic: Okay, I still find it easier to read.

src/macros.jl
```diff
@@ -27,13 +28,52 @@ We can create an affine system as well:
...
+matrix as `I(n)`, for example:
+
+```jldoctest
+julia> @map x -> I(5)*x
+IdentityMap(5)
+```
 """
+macro map(ex, dom)
```
schillic: Do you mean `dim`?

mforets: it was supposed to be `dom`, but then it was `dim` 😄  for cases such as `ConstrainedLinearMap` they will also use this function, so i just kept `dom` for `domain`.

src/macros.jl
mforets: a more generic name for what? the variable `ex`?

schillic: No, `dom` :wink:
"`additional optional arguments`"

mforets: Ah, sure. I can rename this argument `dom -> args`.

src/macros.jl
schillic: There is still the `a the` :wink:

PR #73, opened 2019-02-10
Add ConstrainedResetMap

schillic:


src/maps.jl
```diff
@@ -325,3 +326,39 @@ function apply(m::ResetMap, x)
     end
     return y
 end
+
+"""
+    ConstrainedResetMap
+
+A reset mapwith state constraints of the form:
```
mforets:
```suggestion
A reset map with state constraints of the form:
```

src/maps.jl
```diff
@@ -325,3 +326,39 @@ function apply(m::ResetMap, x)
...
+inputdim(::ConstrainedResetMap) = 0
+outputdim(m::ConstrainedResetMap) = m.dim
+islinear(::ConstrainedResetMap) = false
+isaffine(::ConstrainedResetMap) = true
+
+# convenience constructor for a list of pairs instead of a dictionary
+ConstrainedResetMap(dim::Int, X::ST, args::Pair{Int, <:N}...) where {N, ST} =
+    ConstrainedResetMap(dim, X, Dict{Int, N}(args))
+
+apply(m::ConstrainedResetMap, x) = apply(ResetMap(m.dim, m.dict), x)
```
mforets: While this is correct, it allocates a new reset map. How about using a union type in the existing  `apply` method for `ResetMap`?

schillic: Yes, I was not sure. Should I change this for all `ConstrainedX` types?

src/maps.jl
```diff
@@ -325,3 +326,39 @@ function apply(m::ResetMap, x)
...
+variables are unchanged.
+
+### Fields
+
+- `dim`  -- dimension
+- `X`    -- state constraints
+- `dict` -- dictionary whose keys are the indices of the variables that are
+            reset, and whose values are the new values
+"""
+struct ConstrainedResetMap{N, ST} <: AbstractMap
```
mforets: This is just a whim, but can we keep the same order as for the `ResetMap`? `dim, dict, X` so

```julia
julia> m = ConstrainedResetMap(10, Dict(9 => 0.), X)
```

schillic: Then we cannot offer the convenience constructor.

src/maps.jl
mforets: No, i think that's fine.. since they are one-liners ..

src/maps.jl
mforets: Ah, ok i see. Then we forget my comment..

PR #77, opened 2019-02-11
#74 - Add ispolynomial

### LINKED ISSUE #74: Add ispolynomial trait
(Useful to dispatch on algorithms that are special for polynomial ODEs.)

mforets:
Closes #74.

### DIFF
--- a/src/abstract.jl
+++ b/src/abstract.jl
@@ -88,11 +88,14 @@ For example, polynomial systems can be nonlinear; hence `isaffine` is `false`.
 function isaffine(::AbstractSystem) end
 
 """
-    isalgebraic(s::AbstractSystem)
+    ispolynomial(s::AbstractSystem)
 
 Specifies if the dynamics of system `s` is specified by polynomial equations.
+
+The criterion refers to the *type* information, not the value. Hence, e.g. a
+`LinearContinuousSystem` is not considered to be of polynomial type.
 """
-function isalgebraic(::AbstractSystem) end
+function ispolynomial(::AbstractSystem) end
 
 """
     AbstractMap
@@ -145,4 +148,3 @@ function isaffine(::AbstractMap) end
 Apply the rule specified by the map to the given arguments.
 """
 function apply end
-
--- a/src/abstract.jl
+++ b/src/abstract.jl
@@ -88,11 +88,11 @@ For example, polynomial systems can be nonlinear; hence `isaffine` is `false`.
 function isaffine(::AbstractSystem) end
 
 """
-    ispolynomial(s::AbstractSystem)
+    isalgebraic(s::AbstractSystem)
 
 Specifies if the dynamics of system `s` is specified by polynomial equations.
 """
-function ispolynomial(::AbstractSystem) end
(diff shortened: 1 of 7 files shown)


schillic:
I would say that the semantics of `ispolynomial` should return `true` for a linear system, etc.

blegat:
> I would say that `ispolynomial` should return `true` for a linear system, etc.

Maybe `isalgebraic` then ?

mforets:
yes, `isalgebraic` for [algebraic variety](https://en.wikipedia.org/wiki/Algebraic_variety#Affine_varieties) seems like a better match.

mforets:
This is ready for review.

schillic:
Error [here](https://travis-ci.org/JuliaReach/MathematicalSystems.jl/jobs/502328842#L621):
```
WARNING: both LazySets and MathematicalSystems export "ResetMap"; uses of it in module Main must be qualified
Reset map: Error During Test at /home/travis/build/JuliaReach/MathematicalSystems.jl/test/maps.jl:144
  Got exception outside of a @test
  UndefVarError: ResetMap not defined
  Stacktrace:
   [1] top-level scope at /home/travis/build/JuliaReach/MathematicalSystems.jl/test/maps.jl:145
```

src/abstract.jl
```diff
@@ -87,6 +87,13 @@ For example, polynomial systems can be nonlinear; hence `isaffine` is `false`.
 """
 function isaffine(::AbstractSystem) end
 
+"""
+    isalgebraic(s::AbstractSystem)
+
+Specifies if the dynamics of system `s` is specified by polynomial equations.
```
schillic: Please clarify this definition.

schillic:
I do not fully understand the terminology. When does a system have "algebraic dynamics? Why does a `LinearSystem` not have algebraic dynamics, but a `PolynomialSystem` of degree 1 has it?
[Here](https://math.stackexchange.com/a/187566) there is a Wikipedia link, which redirects to [arithmetic dynamics](https://en.wikipedia.org/wiki/Arithmetic_dynamics). Are you sure this is the right term?

mforets:
> I would say that ispolynomial should return true for a linear system, etc.

`ispolynomialtype` then?

schillic:
Actually we already have this semantic inconsistency (`islinear` returns `false` for `AffineContinuousSystem`s even if the additive dynamics is zero). So I give up and you can go with `ispolynomial` or `isalgebraic`.
But I think that this confusion should be prevented with good documentation that these methods do not express system semantics but rather type categories.

mforets:
Yes, see the 2nd paragraph in the [Notes section of that method](https://juliareach.github.io/MathematicalSystems.jl/latest/lib/methods/#MathematicalSystems.islinear-Tuple{AbstractSystem}).

Perhaps `isfootype` is better, i believe, but that amounts to changing the existing methods, so i that would be a separate PR.

schillic:
> Perhaps isfootype is better, i believe, but that amounts to changing the existing methods, so i that would be a separate PR.

I would not change all names. Just stay consistent and everything is fine.

PR #46, opened 2018-12-17
Update docs script

mforets:
This PR fixes 3 deprecations suggested by `Documenter`:

Current `master`:

```julia
$ julia1 --color=yes docs/make.jl
┌ Warning: `format = :html` is deprecated, use `format = Documenter.HTML()` instead.
│   caller = ip:0x0
└ @ Core :-1
[ Info: SetupBuildDirectory: setting up build directory.
[ Info: ExpandTemplates: expanding markdown templates.
[ Info: CrossReferences: building cross-references.
[ Info: CheckDocument: running document checks.
[ Info: Populate: populating indices.
[ Info: RenderDocument: rendering document.
┌ Warning: the `julia` keyword argument to `Documenter.deploydocs` is removed. Use Travis Build Stages for determining from where to deploy instead. See the section a
bout Hosting in the Documenter manual for more details.
│   caller = ip:0x0
└ @ Core :-1
[ Info: skipping docs deployment.
```

After fixing those two:

```julia
$ julia1 --color=yes docs/make.jl
[ Info: SetupBuildDirectory: setting up build directory.
[ Info: ExpandTemplates: expanding markdown templates.
[ Info: CrossReferences: building cross-references.
[ Info: CheckDocument: running document checks.
[ Info: Populate: populating indices.
[ Info: RenderDocument: rendering document.
┌ Warning: the `osname` keyword argument to `Documenter.deploydocs` is removed. Use Travis Build Stages for determining from where to deploy instead. See the section
about Hosting in the Documenter manual for more details.
│   caller = ip:0x0
└ @ Core :-1
[ Info: skipping docs deployment.
```

After fixing `osname`:

```julia
$ julia1 --color=yes docs/make.jl
[ Info: SetupBuildDirectory: setting up build directory.
[ Info: ExpandTemplates: expanding markdown templates.
[ Info: CrossReferences: building cross-references.
[ Info: CheckDocument: running document checks.
[ Info: Populate: populating indices.
[ Info: RenderDocument: rendering document.
┌ Info: Deployment criteria:
│ - ✔ ENV["TRAVIS_REPO_SLUG"]="" occurs in repo="github.com/JuliaReach/MathematicalSystems.jl.git"
│ - ✘ ENV["TRAVIS_PULL_REQUEST"]="" is "false"
│ - ✔ ENV["TRAVIS_TAG"]="" is (i) empty or (ii) a valid VersionNumber
[...]

PR #81, opened 2019-03-06
Fix deploydocs & clean up Travis script

schillic:
Since #46 the documentation has not been deployed. I think this PR fixes that, together with some minor clean-up.

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -12,7 +12,8 @@ makedocs(
         "Types" => "lib/types.md",
         "Methods" => "lib/methods.md"],
         "About" => "about.md"
-    ]
+    ],
+    strict = true
 )
 
 deploydocs(
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -1,9 +1,9 @@
 using Documenter, MathematicalSystems
 
 makedocs(
-    doctest = true,
     modules = [MathematicalSystems],
-    format = :html,
+    # See https://github.com/JuliaDocs/Documenter.jl/issues/868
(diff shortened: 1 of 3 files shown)


docs/make.jl
```diff
@@ -1,7 +1,7 @@
 using Documenter, MathematicalSystems
 
 makedocs(
-    doctest = true,  # use this flag to skip doctests (saves time!)
+    doctest = true,
```
blegat: Why do we want to skip it ? I would rather add `strict=true` so that CI fails when doctests fail

schillic: I just removed the comment.

schillic: > I would rather add strict=true so that CI fails when doctests fail

Okay, I can add that here.

blegat: What is this flag for ? I would recommend this setup for documenter:
https://github.com/JuliaOpt/JuMP.jl/blob/c01802926c34aa30e5d272ac1063f1595084aa4a/docs/make.jl#L12-L15

schillic: True, we can remove the flag entirely because this is the default value.
I can add the other option.

schillic: To answer your question, it can be used to deactivate doctests (see [here](https://juliadocs.github.io/Documenter.jl/stable/man/doctests/#Skipping-Doctests-1)).

schillic:
Thanks for the comments.

schillic:
Unfortunately, this PR did not help. I created #83.

PR #84, opened 2019-03-12
#78 - Add implicit types

### LINKED ISSUE #78: Add system defined by a right hand side function
To handle the case where the right-hand sides are *not* polynomial, we need to add a new system type. (The example that i have in mind is the [quadrotor model from this benchmark](https://github.com/JuliaReach/ReachabilityBenchmarks/pull/70/files#diff-3ae650d07afc46a0c2d31e6c8bc44dfeR87).)

There are two alternatives:

1- write a more generic system type that can handle other types of terms, such as trigonometric, logarithm, exponential, ...
2- write a system type that can handle a right-hand side defined implicitly through a function `f`.

Perhaps we should do both, but thinking about (2), it could be along these lines:

```julia
struct ImplicitContinuousSystem{FT}
    f::FT  # right-hand side
    statedim::Int
end
```
Other names that come to mind: `BlackBoxContinuousSystem`, `ODEContinuousSystem`, ...

mforets:
Closes #78.

### DIFF
--- a/src/systems.jl
+++ b/src/systems.jl
@@ -778,12 +778,13 @@ for (Z, AZ) in ((:ConstrainedBlackBoxControlContinuousSystem, :AbstractContinuou
         struct $(Z){FT, ST, UT} <: $(AZ)
             f::FT
             statedim::Int
+            inputdim::Int
             X::ST
             U::UT
         end
         statedim(s::$Z) = s.statedim
         stateset(s::$Z) = s.X
-        inputdim(s::$Z) = 0
+        inputdim(s::$Z) = s.inputdim
         inputset(s::$Z) = s.U
         islinear(::$Z) = false
         isaffine(::$Z) = false
@@ -804,6 +805,7 @@ of the form:
 
 - `f`        -- function that holds the right-hand side
 - `statedim` -- number of state variables
+- `inputdim` -- number of input variables
 - `X`        -- state constraints
 - `U`        -- input constraints
 """
@@ -823,6 +825,7 @@ of the form:
 
 - `f`        -- function that holds the right-hand side
 - `statedim` -- number of state variables
+- `inputdim` -- number of input variables
 - `X`        -- state constraints
 - `U`        -- input constraints
 """
--- a/src/systems.jl
+++ b/src/systems.jl
@@ -761,7 +761,7 @@ Discrete-time system defined by a right-hand side with state constraints
 of the form:
 
 ```math
-    x_{k+1} = f(x_k), x(t) ∈ \\mathcal{X}.
+    x_{k+1} = f(x_k), x_k ∈ \\mathcal{X}.
 ```
 
 ### Fields
--- a/src/systems.jl
+++ b/src/systems.jl
@@ -1,5 +1,5 @@
-using MultivariatePolynomials
-using MultivariatePolynomials: AbstractPolynomialLike
+import MultivariatePolynomials
+import MultivariatePolynomials: AbstractPolynomialLike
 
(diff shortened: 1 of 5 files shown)


blegat:
Why the name `Implicit` ? It makes me thing about differential algebraic because you do not define the trajectory explicitly, you need to solver a system.
Why not rather `Nonlinear` ? Polynomial are also non-linear but they are a specific case for which we have a type exploiting the structure. `BlackBox` also sounds good
Maybe `Implicit` is the right name, I am just not sure where it comes from

schillic:
Hm, `Nonlinear` does not nail it down for me. The docstring says `system defined by a right-hand side of the form: x' = f(t)`. `BlackBox` is fine.

mforets:
yeah `Implicit` is not a good name. I am fine with `Nonlinear`. Anyway, `BlackBox` seems to win :)

blegat:
Let's go with BlackBox :)

src/systems.jl
```diff
@@ -673,3 +673,157 @@ Discrete-time polynomial system with state constraints:
...
+ConstrainedBlackBoxContinuousSystem
+
+@doc """
+    ConstrainedBlackBoxDiscreteSystem <: AbstractDiscreteSystem
+
+Discrete-time system defined by a right-hand side with state constraints
+of the form:
+
+```math
+    x_{k+1} = f(x_k), x(t) ∈ \\mathcal{X}.
```
schillic:
```suggestion
    x_{k+1} = f(x_k), x_k ∈ \\mathcal{X}.
```

src/systems.jl
```diff
@@ -673,3 +673,157 @@ Discrete-time polynomial system with state constraints:
...
+        struct $(Z){FT, ST, UT} <: $(AZ)
+            f::FT
+            statedim::Int
+            X::ST
+            U::UT
+        end
+        statedim(s::$Z) = s.statedim
+        stateset(s::$Z) = s.X
+        inputdim(s::$Z) = 0
+        inputset(s::$Z) = s.U
```
schillic: Why `inputdim(s::$Z) = 0`?
And should there not be a second function for the inputs? (`f` replaces `A`, so what replaces `B`?)

src/systems.jl
```diff
@@ -673,3 +673,157 @@ Discrete-time polynomial system with state constraints:
...
+
+- `f`        -- function that holds the right-hand side
+- `statedim` -- number of state variables
+- `X`        -- state constraints
+- `U`        -- input constraints
+"""
+ConstrainedBlackBoxControlContinuousSystem
+
+@doc """
+    ConstrainedBlackBoxControDiscreteSystem <: AbstractDiscreteSystem
```
schillic:
```suggestion
    ConstrainedBlackBoxControlDiscreteSystem <: AbstractDiscreteSystem
```

test/continuous.jl
```diff
@@ -170,3 +170,31 @@ end
...
+    @test dx ≈ [0.0, -1.0]
+end
+
+@testset "Continuous system defined by a function with state constraints" begin
+    function vanderpol!(x, dx)
+        dx[1] = x[2]
+        dx[2] = x[2] * (1-x[1]^2) - x[1]
+        return dx
+    end
+    s = BlackBoxContinuousSystem(vanderpol!, 2, xxxx)
```
schillic: I think you wanted to add something here :smile:

src/systems.jl
mforets: Good point. Actually i think we have to add this information as a new field. The inputs are considered in `f` as well, think of it as: `f : (x(t), u(t)) -> f(x, u)`.

mforets:
Thanks for the review!

PR #89, opened 2019-03-17
Improve multiplication of identity multiple with vec or mat

mforets:


mforets:
Ok, we don't have to pass through the uniform scaling ([see this function](https://github.com/JuliaLang/julia/blob/master/stdlib/LinearAlgebra/src/uniformscaling.jl#L200). The only difference is that here we check that the dimensions match.

PR #94, opened 2019-08-09
Update and fix coverage

schillic:
This also fixes a new namespace conflict with `LazySets` in the tests.
Sorry to tell you that 100% code coverage was just a fairy tale.


.travis.yml
```diff
@@ -16,7 +16,7 @@ matrix:
   - julia: nightly
 
 after_success:
-  - julia -e 'using Pkg; Pkg.add("Coverage"); using Coverage; Coveralls.submit(process_folder()); Codecov.submit(process_folder())'
+  - julia -e 'using Pkg; Pkg.add("Coverage"); using Coverage; Codecov.submit(process_folder())'
```
blegat: We can now just do `codecov: true` at the beginning of the file

schillic: No, that is not sufficient (I tried it). The `Coveralls.submit(process_folder());` causes an error and that prevents the following command.

mforets: Kind of confusing because [see here](https://github.com/JuliaCI/Coverage.jl#tracking-coverage-with-codecovio) it says (point 3): "add `codecov: true`". Then "... you can also add...". I know that you tried the different combinations, and it works now, so `codecov: true` is never needed or it is needed only if we want to use other tools such as `Coveralls`?

schillic: I am also quite confused because in other Julia repositories it seems to work with the old command.

I tried adding `codecov: true` in addition to the old Julia command including `Coveralls.submit(process_folder());`, and this had the same problem (admittedly, I tried in a different Julia repository, but I guess the build process is the same).

What I did not try is removing the Julia command altogether and only adding `codecov: true`. Maybe this is what @blegat meant.

blegat: Yes you can completely remove the `after_success` and only put `codecov: true`

schillic: Any preferences? Should I change it?

blegat: Yes, I think any Julia package should use the new syntax. If you use a custom `after_success`, it seems you are doing something custom which is not exactly `codecov: true` so it's confusing

schillic: Great, now I have to change all our repos again :rofl:

schillic: See #95.

PR #98, opened 2019-11-25
Update doc setup

mforets:


docs/Project.toml
```diff
@@ -2,4 +2,4 @@
 Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
 
 [compat]
-Documenter = "~0.21"
+Documenter = "^0.23"
```
schillic: Does v0.24 not work?

mforets: Well not quite, i have mentioned this earlier in LazySets. The juliareach css is messed up with the new style.

mforets: ... if i recall correctly, the main thing of that css was to align the text. Perhaps by now there is another way to do this in Documenter.

PR #99, opened 2019-11-25
#96 - Successor state for some discrete systems

### LINKED ISSUE #96: Add function to return the successor in a discrete system
`successor(system::ConstrainedLinearControlDiscreteSystem, x, u)` would return the result of `system.A*x + system.B*u` and so on.

mforets:
Closes #96.

### DIFF
--- a/src/successor.jl
+++ b/src/successor.jl
@@ -39,7 +39,7 @@ end
 """
     successor(system::ConstrainedDiscreteIdentitySystem, x::AbstractVector)
 
-Return the successor state of a `ConstrainedDiscreteIdentitySystem`.
+Return the successor state of a `LinearDiscreteSystem`.
 
 ### Input
 
--- a/src/successor.jl
+++ b/src/successor.jl
@@ -43,7 +43,7 @@ Return the successor state of a `ConstrainedDiscreteIdentitySystem`.
 
 ### Input
 
-- `system` -- `ConstrainedDiscreteIdentitySystem`
+- `system` -- `LinearDiscreteSystem`
 - `x`      -- state (it should be any vector type)
 
 ### Output
--- a/src/successor.jl
+++ b/src/successor.jl
@@ -83,7 +83,7 @@ Return the successor state of a `LinearControlDiscreteSystem`.
 
 - `system` -- `LinearControlDiscreteSystem`
 - `x`      -- state (it should be any vector type)
-- `u`      -- state (it should be any vector type)
+- `u`      -- input (it should be any vector type)
 
 ### Output
 
--- a/src/successor.jl
+++ b/src/successor.jl
@@ -1,18 +1,19 @@
 """
-    successor(system::DiscreteIdentitySystem, x)
+    successor(system::DiscreteIdentitySystem, x::AbstractVector)
 
 Return the successor state of a `DiscreteIdentitySystem`.
 
 ### Input
 
 - `system` -- `DiscreteIdentitySystem`
-- `x`      -- state (it can be of any type)
+- `x`      -- state (it should be any vector type)
 
 ### Output
 
 The same state `x`.
 """
-function successor(system::DiscreteIdentitySystem, x)
+function successor(system::DiscreteIdentitySystem, x::AbstractVector)
+    @assert statedim(system) == length(x)
     return x
 end
 
--- a/src/successor.jl
+++ b/src/successor.jl
@@ -1,3 +1,9 @@
+# helper functions to check that the given system can be applied to the states/inputs
+@inline _is_conformable_state(system, x) = statedim(system) == length(x)
+@inline _is_conformable_input(system, u) = inputdim(system) == length(u)
+@inline _in_stateset(system, x) = x ∈ stateset(system)
+@inline _in_inputset(system, u) = u ∈ inputset(system)
+
 """
     successor(system::DiscreteIdentitySystem, x::AbstractVector)
 
@@ -13,7 +19,7 @@ Return the successor state of a `DiscreteIdentitySystem`.
 The same state `x`.
 """
 function successor(system::DiscreteIdentitySystem, x::AbstractVector)
-    @assert statedim(system) == length(x)
+    !_is_conformable_state(system, x) && throw(ArgumentError())
     return x
 end
 
@@ -32,8 +38,7 @@ Return the successor state of a `ConstrainedDiscreteIdentitySystem`.
 The same state `x`.
 """
 function successor(system::ConstrainedDiscreteIdentitySystem, x::AbstractVector)
-    @assert statedim(system) == length(x)
-    @assert x ∈ stateset(system)
+    (!_is_conformable_state(system, x) || !_in_stateset(system, x)) && throw(ArgumentError())
     return x
 end
 
@@ -52,7 +57,7 @@ Return the successor state of a `LinearDiscreteSystem`.
 The result of applying the system to `x`.
 """
 function successor(system::LinearDiscreteSystem, x::AbstractVector)
-    @assert statedim(system) == length(x)
+    !_is_conformable_state(system, x) && throw(ArgumentError())
     return system.A * x
 end
 
@@ -71,7 +76,7 @@ Return the successor state of a `AffineDiscreteSystem`.
 The result of applying the system to `x`.
 """
 function successor(system::AffineDiscreteSystem, x::AbstractVector)
-    @assert statedim(system) == length(x)
+    !_is_conformable_state(system, x) && throw(ArgumentError())
     return system.A * x + system.b
 end
 
@@ -91,8 +96,7 @@ Return the successor state of a `LinearControlDiscreteSystem`.
 The result of applying the system to `x`, with input `u`.
 """
 function successor(system::LinearControlDiscreteSystem, x::AbstractVector, u::AbstractVector)
-    @assert statedim(system) == length(x)
-    @assert inputdim(system) == length(u)
+    (!_is_conformable_state(system, x) || !_is_conformable_input(system, u)) && throw(ArgumentError())
     return system.A * x + system.B * u
 end
 
@@ -111,8 +115,7 @@ Return the successor state of a `ConstrainedLinearDiscreteSystem`.
 The result of applying the system to `x`.
 """
 function successor(system::ConstrainedLinearDiscreteSystem, x::AbstractVector)
-    @assert statedim(system) == length(x)
-    @assert x ∈ stateset(system)
+    (!_is_conformable_state(system, x) || !_in_stateset(system, x)) && throw(ArgumentError())
     return system.A * x
 end
 
@@ -131,8 +134,7 @@ Return the successor state of a `ConstrainedAffineDiscreteSystem`.
 The result of applying the system to `x`.
 """
 function successor(system::ConstrainedAffineDiscreteSystem, x)
-    @assert statedim(system) == length(x)
-    @assert x ∈ stateset(system)
+    (!_conformable(system, x) || !_in_stateset(system, x)) && throw(ArgumentError())
     return system.A * x + system.b
 end
 
@@ -152,10 +154,8 @@ Return the successor state of a `ConstrainedLinearControlDiscreteSystem`.
 The result of applying the system to `x`, with input `u`.
 """
 function successor(system::ConstrainedLinearControlDiscreteSystem, x::AbstractVector, u::AbstractVector)
-    @assert statedim(system) == length(x)
-    @assert x ∈ stateset(system)
-    @assert inputdim(system) == length(u)
-    @assert u ∈ inputset(system)
+    (!_is_conformable_state(system, x) || !_is_conformable_input(system, u) ||
+    !_in_stateset(system, x) || !_in_inputset(system, u)) && throw(ArgumentError())
     return system.A * x + system.B * u
 end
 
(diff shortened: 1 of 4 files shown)


src/successor.jl
```diff
@@ -0,0 +1,221 @@
...
+- `system` -- `ConstrainedBlackBoxDiscreteSystem`
+- `x`      -- state (it should be any vector type)
+
+### Output
+
+The result of applying the system to `x`.
+"""
+function successor(system::ConstrainedBlackBoxDiscreteSystem, x::AbstractVector)
+    @assert statedim(system) == length(x)
+    @assert x ∈ system.X
```
ueliwechsler: For consistency you could have here as above `x ∈ stateset(system)` instead of `x ∈ system.X`.

mforets: Thanks for the comment, i'll change this.

mforets:
```suggestion
    @assert x ∈ stateset(system)
```

src/successor.jl
```diff
@@ -0,0 +1,221 @@
...
+- `system` -- `ConstrainedAffineDiscreteSystem`
+- `x`      -- state (it should be any vector type)
+
+### Output
+
+The result of applying the system to `x`.
+"""
+function successor(system::ConstrainedAffineDiscreteSystem, x)
+    @assert statedim(system) == length(x)
+    @assert x ∈ system.X
```
mforets:
```suggestion
    @assert x ∈ stateset(system)
```

mforets:
Any other comments before merging?

src/successor.jl
```diff
@@ -0,0 +1,221 @@
...
+- `system` -- `ConstrainedDiscreteIdentitySystem`
+- `x`      -- state (it should be any vector type)
+
+### Output
+
+The same state `x`.
+"""
+function successor(system::ConstrainedDiscreteIdentitySystem, x::AbstractVector)
+    @assert statedim(system) == length(x)
+    @assert x ∈ stateset(system)
```
blegat: I would add errors instead of asserts. Asserts may be disabled so they should be reserved to errors that you can certify will not happen if there is no bug in the package. Here it could error if the user give a bad argument so it should be an `ArgumentError` instead.

schillic: I agree that it should be an `ArgumentError`, but I also like to be able to disable these checks in performance-critical code...

src/successor.jl
```diff
@@ -0,0 +1,221 @@
...
+
+- `system` -- `DiscreteIdentitySystem`
+- `x`      -- state (it can be of any type)
+
+### Output
+
+The same state `x`.
+"""
+function successor(system::DiscreteIdentitySystem, x)
+    return x
```
schillic: You can `@assert` (or whatever) `statedim(system) == length(x)` here as well.

src/successor.jl
```diff
@@ -0,0 +1,221 @@
...
+The same state `x`.
+"""
+function successor(system::ConstrainedDiscreteIdentitySystem, x::AbstractVector)
+    @assert statedim(system) == length(x)
+    @assert x ∈ stateset(system)
+    return x
+end
+
+"""
+    successor(system::ConstrainedDiscreteIdentitySystem, x::AbstractVector)
```
schillic:
```suggestion
    successor(system::LinearDiscreteSystem, x::AbstractVector)
```

src/successor.jl
```diff
@@ -0,0 +1,221 @@
...
+
+Return the successor state of a `ConstrainedDiscreteIdentitySystem`.
```
schillic:
```suggestion
Return the successor state of a `LinearDiscreteSystem`.
```

src/successor.jl
```diff
@@ -0,0 +1,221 @@
...
+
+### Input
+
+- `system` -- `ConstrainedDiscreteIdentitySystem`
```
schillic:
```suggestion
- `system` -- `LinearDiscreteSystem`
```

src/successor.jl
```diff
@@ -0,0 +1,221 @@
...
+"""
+    successor(system::LinearControlDiscreteSystem, x::AbstractVector, u::AbstractVector)
+
+Return the successor state of a `LinearControlDiscreteSystem`.
+
+### Input
+
+- `system` -- `LinearControlDiscreteSystem`
+- `x`      -- state (it should be any vector type)
+- `u`      -- state (it should be any vector type)
```
schillic:
```suggestion
- `u`      -- input (it should be any vector type)
```

src/successor.jl
```diff
@@ -0,0 +1,221 @@
...
+"""
+    successor(system::ConstrainedLinearControlDiscreteSystem, x::AbstractVector, u::AbstractVector)
+
+Return the successor state of a `ConstrainedLinearControlDiscreteSystem`.
+
+### Input
+
+- `system` -- `ConstrainedLinearControlDiscreteSystem`
+- `x`      -- state (it should be any vector type)
+- `u`      -- state (it should be any vector type)
```
schillic:
```suggestion
- `u`      -- input (it should be any vector type)
```

src/successor.jl
```diff
@@ -0,0 +1,221 @@
...
+"""
+    successor(system::ConstrainedAffineControlDiscreteSystem, x::AbstractVector, u::AbstractVector)
+
+Return the successor state of a `ConstrainedAffineControlDiscreteSystem`.
+
+### Input
+
+- `system` -- `ConstrainedAffineControlDiscreteSystem`
+- `x`      -- state (it should be any vector type)
+- `u`      -- state (it should be any vector type)
```
schillic:
```suggestion
- `u`      -- input (it should be any vector type)
```

src/successor.jl
mforets: OK i went for the error code because of the nature of these checks. About disabling them, it is possible to do so by overriding the helper functions as the following example shows:
 
```julia
using Revise, MathematicalSystems, LazySets, BenchmarkTools

A = [1. 1; 1 -1]
B = Matrix([0.5 1.5]')
X = Hyperrectangle([0.0, 0.3], [2.0, 2.0])
U = Interval(0.0, 1.0)
sys = ConstrainedLinearControlDiscreteSystem(A, B, X, U)
x = [0.0; 0.3]
u = [0.5]

julia> @btime successor($sys, $x, $u)
  141.346 ns (3 allocations: 288 bytes)
2-element Array{Float64,1}:
 0.55
 0.45
```

Now disable the membership checks:

```julia
julia> MathematicalSystems._in_stateset(system, x) = true

julia> MathematicalSystems._in_inputset(system, u) = true

julia> @btime successor($sys, $x, $u)
  133.130 ns (3 allocations: 288 bytes)
2-element Array{Float64,1}:
 0.55
 0.45
```
Now disable the length checks:
```julia
julia> MathematicalSystems._is_conformable_state(system, x) = true

julia> MathematicalSystems._is_conformable_input(system, u) = true

julia> @btime successor($sys, $x, $u)
  132.626 ns (3 allocations: 288 bytes)
2-element Array{Float64,1}:
 0.55
 0.45
 ```

File: src/successor.jl [hunk 11]
```diff
@@ -0,0 +1,219 @@
...
+
+- `system` -- `ConstrainedAffineDiscreteSystem`
+- `x`      -- state (it should be any vector type)
+
+### Output
+
+The result of applying the system to `x`.
+"""
+function successor(system::ConstrainedAffineDiscreteSystem, x)
+    (!_conformable(system, x) || !_in_stateset(system, x)) && throw(ArgumentError())
```
ueliwechsler: `_is_conformable_state(system, x)` instead of  `!_conformable(system, x)`

mforets: Thanks for pointing this out.

mforets:
```suggestion
    (!_is_conformable_state(system, x) || !_in_stateset(system, x)) && throw(ArgumentError())
```

src/successor.jl
```diff
@@ -0,0 +1,219 @@
...
+- `x`      -- state (it should be any vector type)
+- `u`      -- input (it should be any vector type)
+
+### Output
+
+The result of applying the system to `x`, with input `u`.
+"""
+function successor(system::ConstrainedLinearControlDiscreteSystem, x::AbstractVector, u::AbstractVector)
+    (!_is_conformable_state(system, x) || !_is_conformable_input(system, u) ||
+    !_in_stateset(system, x) || !_in_inputset(system, u)) && throw(ArgumentError())
```
blegat: I don't like the idea of having to use type piracy to disable the check. Why not use a keyword argument instead ?

ueliwechsler: What about having a macro such as `@inbound` in order to enforce fast execution without checks?  Would this be difficult to implement?
i.e.  `x⁺ = @comformable successor(sys, x,u)`

mforets: There is the package https://github.com/jw3126/ArgCheck.jl, perhaps useful in this case.

ueliwechsler: `ArgCheck.jl` looks neat. But I don't think it enables you to omit the checks to increase performances.

schillic: I like the macro idea, but keyword args seem easier to add.

ueliwechsler: What about we add keyword arguments now and then later define a macro that just adds the keyword arguments to the function call? Would this be a reasonable thing to do?

blegat: > What about we add keyword arguments now and then later define a macro that just adds the keyword arguments to the function call? Would this be a reasonable thing to do?

Yes, that would be reasonable. A macro should not hide a bad low level interface but it can improve usability of a good low level interface

ueliwechsler: If we add the keyword arguments, I could have a go at the macro. I haven't done a lot of metaprogramming so far, it would be a nice way to extend my knowledge.

I have written a quick macro as a first draft (see below). I am not really used to metaprogramming, so if you have any inputs on how I could improve this or where I have to be careful, this would be great. But I guess, we can discuss this in a new issue.

```julia
using MathematicalSystems, LazySets, BenchmarkTools

@inline _is_conformable_state(system, x) = statedim(system) == length(x)
@inline _is_conformable_input(system, u) = inputdim(system) == length(u)
@inline _in_stateset(system, x) = x ∈ stateset(system)
@inline _in_inputset(system, u) = u ∈ inputset(system)

function successor(system::LinearDiscreteSystem, x::AbstractVector; check=true)
    check && !_is_conformable_state(system, x) && throw(ArgumentError("invalid arguments"))
    return system.A * x
end

function successor(system::ConstrainedLinearControlDiscreteSystem, x::AbstractVector, u::AbstractVector; check=true)
    check && (!_is_conformable_state(system, x) || !_is_conformable_input(system, u) || !_in_stateset(system, x) || !_in_inputset(system, u)) && throw(ArgumentError("invalid arguments"))
    return system.A * x + system.B * u
end

macro comformable(expr)
    func = expr.args[1]
    kw1 = Expr(:kw, :check, false)
    vals = []
    for vl in expr.args[2:end]
        if !(typeof(vl) <: Expr)
            push!(vals, vl)
        end
    end
    func2 = Expr(:call, func, vals..., kw1 )
    return esc(func2)
end

A = [1. 1; 1 -1]
B = Matrix([0.5 1.5]')
X = Hyperrectangle([0.0, 0.3], [2.0, 2.0])
U = Interval(0.0, 1.0)
sys = ConstrainedLinearControlDiscreteSystem(A, B, X, U)
x = [0.0; 0.3]
u = [0.5]

x1 = @btime successor($sys, $x, $u; check=true)
#  198.276 ns (3 allocations: 288 bytes)
x2 = @btime @comformable successor($sys, $x, $u)
# 187.421 ns (3 allocations: 288 bytes)
```

And the corresponding error messages:
```julia
x = [0.0; 0.3; 0.3] # has wrong dimensions
u = [0.5]
x1 = successor(sys, x, u; check=true)
# ERROR: ArgumentError: invalid arguments
x2 = @comformable successor(sys, x, u)
# ERROR: DimensionMismatch("second dimension of A, 2, does not match length of x, 3")
```

mforets: Thanks, i've outsourced the last comment to a new issue, see https://github.com/JuliaReach/MathematicalSystems.jl/issues/101

mforets:
This PR is again ready for review. I've added a keyword argument `check_invariant` to make the membership checks if the system has state or input constraints.

I left the dimensional checks to be mandatory, as it is already the case for the constructors, and because their associated runtime is negligible compared to the matrix-vector computation and the invariance check.

src/successor.jl
```diff
@@ -0,0 +1,259 @@
...
+                       the state set
+
+### Output
+
+The result of applying the system to `x`.
+"""
+function successor(system::ConstrainedLinearDiscreteSystem, x::AbstractVector;
+                   check_invariant::Bool=true)
+    !_is_conformable_state(system, x) && throw(ArgumentError())
+    if check_invariant
```
blegat: Wouldn't `check_constraints` be more consistent since we call these sets the constraint sets?

mforets: Yes. I've changed the keyword  argument to `check_constraints`.

src/successor.jl
```diff
@@ -109,105 +109,105 @@ end
...
-- `system`          -- `ConstrainedLinearControlDiscreteSystem`
-- `x`               -- state (it should be any vector type)
-- `u`               -- input (it should be any vector type)
-- `check_invariant` -- (optional, default: `true`) check if the state belongs to
-                       the state set
+- `system`            -- `ConstrainedLinearControlDiscreteSystem`
+- `x`                 -- state (it should be any vector type)
+- `u`                 -- input (it should be any vector type)
+- `check_constraints` -- (optional, default: `true`) check if the state belongs to
+                         the state set
```
schillic: It also checks for the input, not only the state.

mforets: Yes. Thanks for double checking the docstring.

mforets:
Thanks for your reviews!

PR #100, opened 2019-11-28
#97 Add some NoisyXYZSystems

### LINKED ISSUE #97: Add NoisyXYZsystem
I would like to add noisy mathematical systems, i.e. a system with additive bounded noise such as `x+ = Ax + Bu + c + Dw`, for the application in robust control.
As an example, I have a version of a `NoisyConstrainedAffineControlDiscreteSystem`. 
First of all, do you agree to include these types? And if so, which system types would need to be extended? For my use-case I am mostly in need of linear and affine ControlDiscreteSystems, however, I assume this should be done for more types.

```julia
using MathematicalSystems

for (Z, AZ) in ((:NoisyConstrainedAffineControlContinuousSystem, :AbstractContinuousSystem),
                (:NoisyConstrainedAffineControlDiscreteSystem, :AbstractDiscreteSystem))
    @eval begin
        struct $(Z){T, MTA <: AbstractMatrix{T}, MTB <: AbstractMatrix{T}, MTD <: AbstractMatrix{T}, VT <: AbstractVector{T}, ST, UT, WT} <: $(AZ)
            A::MTA
            B::MTB
            c::VT
            D::MTD
            X::ST
            U::UT
            W::WT
            function $(Z)(A::MTA, B::MTB, c::VT, X::ST, U::UT) where {T, MTA <: AbstractMatrix{T}, MTB <: AbstractMatrix{T}, VT <: AbstractVector{T}, ST, UT}
                @assert checksquare(A) == length(c) == size(B, 1) == size(W,1)
                return new{T, MTA, MTB, VT, MTD, ST, UT, WT}(A, B, c, D, X, U, W)
            end
        end
        statedim(s::$Z) = length(s.c)
        stateset(s::$Z) = s.X
        inputdim(s::$Z) = size(s.B, 2)
        inputset(s::$Z) = s.U
        noisedim(s::$Z) = size(s.D, 2)
        noiseset(s::$Z) = s.W
        islinear(::$Z) = false
        isaffine(::$Z) = true
        isnoisy(::$Z) = true
    end
end

### PR DESCRIPTION
**Pull request by ueliwechsler**
Closes #97 

I added the types:
- `NoisyConstrainedLinearDiscreteSystem`, `NoisyConstrainedLinearContinuousSystem`,
- `NoisyConstrainedLinearControlDiscreteSystem`, `NoisyConstrainedLinearControlContinuousSystem`
- `NoisyConstrainedAffineControlDiscreteSystem`, `NoisyConstrainedBlackBoxControlContinuousSystem`
- `NoisyConstrainedBlackBoxControlDiscreteSystem`, `NoisyConstrainedAffineControlContinuousSystem`

For the checks, I'd like to use a `dim(X)` function, which I could not make work. I guess I have to wait for `MathematicalSets.jl` to do so ;)  

Additional to the default constructor `NoisyXYZSystem(A,B,c,D,X,U,W)`, I added a constructor which does construct `D=Identymatrix` automatically by `NoisyXYZSystem(A,B,c,X,U,W)`.

### DIFF
--- a/docs/src/lib/types.md
+++ b/docs/src/lib/types.md
@@ -43,8 +43,8 @@ ConstrainedBlackBoxContinuousSystem
 ConstrainedBlackBoxControlContinuousSystem
 NoisyConstrainedLinearContinuousSystem
 NoisyConstrainedLinearControlContinuousSystem
-NoisyConstrainedBlackBoxControlContinuousSystem
 NoisyConstrainedAffineControlContinuousSystem
+NoisyConstrainedBlackBoxControlContinuousSystem
 ```
 
 ## Discrete Systems
--- a/docs/src/lib/types.md
+++ b/docs/src/lib/types.md
@@ -41,6 +41,10 @@ ConstrainedPolynomialContinuousSystem
 BlackBoxContinuousSystem
 ConstrainedBlackBoxContinuousSystem
 ConstrainedBlackBoxControlContinuousSystem
+NoisyConstrainedLinearContinuousSystem
+NoisyConstrainedLinearControlContinuousSystem
+NoisyConstrainedBlackBoxControlContinuousSystem
+NoisyConstrainedAffineControlContinuousSystem
 ```
 
 ## Discrete Systems
@@ -62,6 +66,10 @@ ConstrainedPolynomialDiscreteSystem
 BlackBoxDiscreteSystem
 ConstrainedBlackBoxDiscreteSystem
 ConstrainedBlackBoxControlDiscreteSystem
+NoisyConstrainedLinearDiscreteSystem
+NoisyConstrainedLinearControlDiscreteSystem
+NoisyConstrainedAffineControlDiscreteSystem
+NoisyConstrainedBlackBoxControlDiscreteSystem
 ```
 
 ## Identity operator
--- a/src/systems.jl
+++ b/src/systems.jl
@@ -344,7 +344,7 @@ for (Z, AZ) in ((:ConstrainedAffineControlContinuousSystem, :AbstractContinuousS
             X::ST
             U::UT
             function $(Z)(A::MTA, B::MTB, c::VT, X::ST, U::UT) where {T, MTA <: AbstractMatrix{T}, MTB <: AbstractMatrix{T}, VT <: AbstractVector{T}, ST, UT}
-                @assert checksquare(A) == length(c) == size(B, 1) 
+                @assert checksquare(A) == length(c) == size(B, 1)
                 return new{T, MTA, MTB, VT, ST, UT}(A, B, c, X, U)
             end
         end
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

File: src/systems.jl [hunk 1]
```diff
@@ -843,3 +843,278 @@ of the form:
...
+for (Z, AZ) in ((:NoisyConstrainedLinearContinuousSystem, :AbstractContinuousSystem),
+                (:NoisyConstrainedLinearDiscreteSystem, :AbstractDiscreteSystem))
+    @eval begin
+        struct $(Z){T, MTA <: AbstractMatrix{T}, MTD <: AbstractMatrix{T}, ST, WT} <: $(AZ)
+            A::MTA
+            D::MTD
+            X::ST
+            W::WT
+            function $(Z)(A::MTA, D::MTD, X::ST, W::WT) where {T, MTA <: AbstractMatrix{T}, MTD <: AbstractMatrix{T}, ST, WT}
+                @assert checksquare(A) == size(D,1) # == dim(X)
```
schillic: I suggest to remove these commented code parts.

src/systems.jl
```diff
@@ -843,3 +843,278 @@ of the form:
...
+                return new{T, MTA, MTD, ST, WT}(A, D, X, W)
+            end
+            function $(Z)(A::MTA, X::ST, W::WT) where {T, MTA <: AbstractMatrix{T}, ST, WT}
+                # @assert checksquare(A) == dim(X)
+                n = size(A,1)
+                D = Matrix{T}(I, n, n)
```
schillic: Maybe create `Diagonal`s instead.
And I think you actually may want to have the dimension of `W` as one entry (but this is not available as you described). From that I am not sure that this is a good default constructor.

src/systems.jl
```diff
@@ -843,3 +843,278 @@ of the form:
...
+                return new{T, MTA, MTA, ST, WT}(A, D, X, W)
+            end
+        end
+        statedim(s::$Z) = checksquare(s.A)
```
schillic: I whould check for a square matrix only in the constructor.

test/continuous.jl
```diff
@@ -217,3 +217,88 @@ end
...
+    W =  BallInf(zeros(l), 1.0)
+    s = NoisyConstrainedBlackBoxControlContinuousSystem(f, n, m, l, X, U, W)
+    @test statedim(s) == n
+    @test s.f == f
+    @test stateset(s) == X
+    @test noisedim(s) == l == dim(W)
+    @test noiseset(s) == W
+    @test inputdim(s) == dim(U)
+    @test !islinear(s) && !isaffine(s) && isnoisy(s)
+    # s = NoisyConstrainedBlackBoxControlContinuousSystem(f, X, U, W)
```
schillic: What about this line?

docs/src/lib/types.md
```diff
@@ -41,6 +41,10 @@ ConstrainedPolynomialContinuousSystem
 BlackBoxContinuousSystem
 ConstrainedBlackBoxContinuousSystem
 ConstrainedBlackBoxControlContinuousSystem
+NoisyConstrainedLinearContinuousSystem
+NoisyConstrainedLinearControlContinuousSystem
+NoisyConstrainedBlackBoxControlContinuousSystem
+NoisyConstrainedAffineControlContinuousSystem
```
schillic: maybe swap these for consistency

ueliwechsler: Done

PR #102, opened 2019-12-02
Update Project.toml

mforets:


Project.toml
```diff
@@ -1,16 +1,16 @@
...
 [deps]
 Espresso = "6912e4f1-e036-58b0-9138-08d1e6358ea9"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 MultivariatePolynomials = "102ac46a-7ee4-5c85-9060-abc95bfdeaa3"
 
 [compat]
-julia = "≥ 1.0.0"
-Espresso = "≥ 0.6.0"
-MultivariatePolynomials = "≥ 0.1.4"
+julia = "1.0, 1.1, 1.2, 1.3"
```
blegat: I would just write `julia = "1"`. The Julia devs are responsible to make this package not break with v1.4, v1.5, ... until v2. It's better to make it installable on any v1.x then so that they automatically test new Julia releases with this package.

mforets:
```suggestion
julia = "1"
```

mforets:
@JuliaRegistrator register()

PR #111, opened 2019-12-11
#110 - Missing square matrix assertions

### LINKED ISSUE #110: Assert square matrix in LinearContinuousSystem constructor

mforets:
Closes #110.

### DIFF
--- a/src/systems.jl
+++ b/src/systems.jl
@@ -1,6 +1,11 @@
 import MultivariatePolynomials
 import MultivariatePolynomials: AbstractPolynomialLike
 
+# check if a matrix is square
+@inline function issquare(A::AbstractMatrix)
+    return size(A, 1) == size(A, 2)
+end
+
 for (Z, AZ) in ((:ContinuousIdentitySystem, :AbstractContinuousSystem),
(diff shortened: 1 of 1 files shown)


src/systems.jl
```diff
@@ -249,8 +258,12 @@ for (Z, AZ) in ((:ConstrainedLinearContinuousSystem, :AbstractContinuousSystem),
         struct $(Z){T, MT <: AbstractMatrix{T}, ST} <: $(AZ)
             A::MT
             X::ST
+            function $(Z)(A::MT, X::ST) where {T, MT <: AbstractMatrix{T}, ST}
+                @assert issquare(A)
```
blegat: Wouldn't an `ArgumentError` be more appropriate than an `@assert` ? This is user-facing so we cannot guarantee that this will never happen

mforets: True, you also raised this issue in the `successor` PR and we agreed to use `ArgumentError`. 

The other existing inner constructors also throw assertions. I agree to change all the inner constructors to throw an `AssertionError`, but i would prefer to do it in a separate PR (after merging this one).

blegat: Sounds good

PR #114, opened 2019-12-13
fix at @map macro

### LINKED ISSUE #113: Map fails if A is a variable
```julia
julia> @map x -> rand(2, 2) * x  # ok
MathematicalSystems.LinearMap{Float64,Array{Float64,2}}([0.8889919743522887 0.1607072511950136; 0.5439088199011097 0.5514228069532758])

julia> A = rand(2, 2);

julia> @map x -> A*x # fails
UndefVarError: A not defined

Stacktrace:
 [1] top-level scope
 [2] eval at ./boot.jl:330 [inlined]
 [3] eval(::Symbol) at /home/mforets/.julia/dev/MathematicalSystems/src/MathematicalSystems.jl:2
 [4] top-level scope at /home/mforets/.julia/dev/MathematicalSystems/src/macros.jl:81
 [5] top-level scope at In[36]:2
```

ueliwechsler:
Closes #113


src/macros.jl
```diff
@@ -61,31 +61,30 @@ macro map(ex, args)
...
-
-        throw(ArgumentError("unable to match the given expression to a known map type"))
-    end
+    local x = (ex.args)[1]
+    local rhs = (ex.args)[2].args[2]
+
+    # x -> I(n)*x
+    # (this rule is more specific than x -> Ax so it should come before it)
+    MT = IdentityMap
+    pat = Meta.parse("I(_n) * $x")
```
mforets: This is not related to the PR fix but while we are at it, should we also make `pat` a `local` variable here, as in line 78? IIUC varaibles should be declared `local` to not be confused with variables of the same name in the calling site.

blegat: This is not part of the ecpression returned by the macro, you can see it with `@macroexpand`

ueliwechsler:
From my side, this would be all.

ueliwechsler:
Sure! Should I add the changes to comply with the source code guidelines?

mforets:
Sure, you can add a new commit to this branch if you prefer, before we merger, or on a future PR.

ueliwechsler:
From my side, this is it.

mforets:
Thanks a lot!

PR #121, opened 2019-12-17
Equality check

### LINKED ISSUE #119: Add Equality operator
To utilize some unit tests, I defined an equality operator for a struct `sys::AbstractSystem` as
```julia

import Base.==
import Base.≈

export _corresponding_type, discretize

function ==(sys1::AbstractSystem, sys2::AbstractSystem)
    if typeof(sys1)!== typeof(sys2)
        return false
    end
    for field in fieldnames(typeof(sys1))
        if getfield(sys1, field) != getfield(sys2, field)
            return false
        end
    end
    return true
end

function ≈(sys1::AbstractSystem, sys2::AbstractSystem)
    if typeof(sys1)!== typeof(sys2)
        return false
    end
    for field in fieldnames(typeof(sys1))
        if !(getfield(sys1, field) ≈ getfield(sys2, field))
            return false
        end
    end
    return true
end
```

What is your opinion on adding the `==` and ` ≈` operator to the package?

ueliwechsler:
closes #119


test/utilities.jl
```diff
@@ -0,0 +1,28 @@
...
+    @test !(s1 ≈ s2)
+
+    x0 = [1.0, 3.0]
+    ivp1 = p = InitialValueProblem(s1, x0)
+    ivp2 = p = InitialValueProblem(s3, x0)
+
+    @test ivp1 == ivp1
+    @test ivp1 != ivp2
+    @test ivp1 ≈ ivp1
+    @test ivp1 ≈ ivp2
```
schillic: Same here:
```julia
     x0 = [1.0, 3.0]
    ivp1 = p = InitialValueProblem(s1, x0)
    ivp2 = p = InitialValueProblem(s2, x0)
    ivp3 = p = InitialValueProblem(s3, x0)
    ivp4 = p = InitialValueProblem(s4, x0)

    @test ivp1 == ivp1
    @test ivp1 != ivp2
    @test ivp1 != ivp3
    @test ivp1 != ivp4
    @test ivp1 ≈ ivp1
    @test !(ivp1 ≈ ivp2)
    @test ivp1 ≈ ivp3
    @test !(ivp1 ≈ ivp4)
```

blegat:
`==` does not redirect to `isequal` but to `===`. `isequal` and `==` are different:
```julia
julia> 0.0 == -0.0
true

julia> isequal(0.0, -0.0)
false
```

src/utilities.jl
```diff
@@ -0,0 +1,25 @@
+import Base: ==
+
+function ==(sys1::AbstractSystem, sys2::AbstractSystem)
+    if typeof(sys1) != typeof(sys2)
+        return false
+    end
+    for field in fieldnames(typeof(sys1))
+        if !(getfield(sys1, field) == getfield(sys2, field))
```
schillic: You did it again :smile:
```suggestion
        if getfield(sys1, field) != getfield(sys2, field)
```

ueliwechsler: I did not change it here because for me it looks more consistent with the `isapprox` method this way and also to directly show that there is a recursion of `==` if a `field` of `sys` is again of type `AbstractSystem`. But I don't have a strong opinion about this.
What is the reason you would change it? More precise code, consistent use of `!=`?

schillic: I also do not have a strong opinion, but I find it easier to read.
Compare `1 ≠ 2` with `¬(1 = 2)`.

PR #122, opened 2019-12-22
Add discretize method for AbstractContinuousSystems

### LINKED ISSUE #107: Add discretize methods for AbstractContinous system
I was playing around with the idea of adding a method that discretizes an `AbstractContinuousSystem` by using either the exact discretization or an Euler discretization if `rank(A) != size(A,1)`, i.e. `A` not invertible.

The method signature would look like this:
`discretize(sys::AbstractContinuousSystem, ΔT::Real)::AbsttractDiscreteSystem`

ueliwechsler:
Again a larger PR 😅; closes #107 

- Introduces helper-method`_complementary_type` to get continuos from discrete and discrete from continuous type.
- adds `discretize(system::AbstractContinuousSystem, ΔT)` method which dispatches to the internal method `_discretize`
- adds dependency on `SparseArrays` (for `spzeros`) 
- adds dependency for Test on `InteractiveUtils`  (for `subtypes`)


src/discretize.jl
```diff
@@ -0,0 +1,173 @@
...
+                D::AbstractMatrix, ΔT::Real; algorithm=:exact)
+
+    Implementation of the discretization algorithm used in `discretize`.
+
+    See [`discretize`](@ref) for more details.
+"""
+function _discretize(A::AbstractMatrix,
+                     B::AbstractMatrix,
+                     c::AbstractVector,
+                     D::AbstractMatrix, ΔT::Real; algorithm=:exact)
```
schillic: Instead of all these explicit copy implementations you could just make `ΔT` the first argument and define default values for the other arguments. Julia will do the rest for you under the hood.

test/discretize.jl
```diff
@@ -0,0 +1,46 @@
+@testset "Convert Continuous to Discrete Type" begin
+    DTYPES = subtypes(AbstractDiscreteSystem)
+    for dtype in DTYPES
+        ctype =  eval.(Meta.parse.(replace(string(dtype), "Discrete" => "Continuous")))
```
schillic:
```suggestion
        ctype = eval.(Meta.parse.(replace(string(dtype), "Discrete" => "Continuous")))
```

test/discretize.jl
```diff
@@ -0,0 +1,46 @@
...
+        @test MathematicalSystems._complementary_type(dtype) == ctype
+    end
+
+    CTYPES = subtypes(AbstractContinuousSystem)
+    for ctype in CTYPES
+        dtype =  eval.(Meta.parse.(replace(string(ctype), "Continuous" => "Discrete")))
```
schillic:
```suggestion
        dtype = eval.(Meta.parse.(replace(string(ctype), "Continuous" => "Discrete")))
```

test/discretize.jl
```diff
@@ -0,0 +1,46 @@
+@testset "Convert Continuous to Discrete Type" begin
+    DTYPES = subtypes(AbstractDiscreteSystem)
+    for dtype in DTYPES
```
schillic: You can merge these lines (same for `CTYPES` below).

test/discretize.jl
```diff
@@ -0,0 +1,46 @@
...
+        dtype =  eval.(Meta.parse.(replace(string(ctype), "Continuous" => "Discrete")))
+        @test MathematicalSystems._complementary_type(ctype) == dtype
+    end
+end
+
+@testset "Exact Discretization of Continous Systems" begin
+    ΔT = 0.1
+    A = [0.5 1; 0.0 0.5]
+    B = Matrix([0.5 1.0]')
+    b = [1.0; 1.0] #because Ax + b
```
schillic: I do not understand the comment

test/discretize.jl
```diff
@@ -0,0 +1,46 @@
...
+    B_d = inv(A)*(A_d - I)*B
+    c_d = inv(A)*(A_d - I)*c
+    b_d = inv(A)*(A_d - I)*b
+    D_d = inv(A)*(A_d - I)*D
+    dict = Dict([:A => A, :B => B, :b => b, :c => c, :D => D,
+                 :X => X, :U => U, :W => W])
+    dict_discretized = Dict([:A => A_d, :B => B_d, :b => b_d, :c => c_d, :D => D_d,
+                 :X => X, :U => U, :W => W])
+    # get affine ctypes
+    CTYPES = filter(x -> (occursin("Linear", string(x)) || occursin("Affine", string(x))) &&
```
schillic: It is sad that we do not have `isaffine(s)` for system types because `(occursin("Linear", string(x)) || occursin("Affine", string(x)))` is just a workaround.

test/discretize.jl
```diff
@@ -0,0 +1,46 @@
...
+                         !occursin("Algebraic", string(x)) , subtypes(AbstractContinuousSystem))
+    DTYPES = MathematicalSystems._complementary_type.(CTYPES)
+    CFIELDS = fieldnames.(CTYPES)
+    CValues = [getindex.(Ref(dict), type) for type in CFIELDS]
+    DValues = [getindex.(Ref(dict_discretized), type) for type in CFIELDS]
+    discretized_function = [discretize(CTYPES[i](CValues[i]...), ΔT) for i=1:length(CTYPES)]
+    discretized_constructed = [DTYPES[i](DValues[i]...) for i=1:length(DTYPES)]
+
+    @test all(discretized_constructed .== discretized_function)
+    @test all(discretized_constructed .≈ discretized_function)
```
schillic: Is this test not implied by the line above?

src/discretize.jl
```diff
@@ -0,0 +1,173 @@
...
+    ### Input
+
+    - `system_type` -- type of `AbstractSystem`
+
+    ### Ouput
+
+    Return complementary type of `system_type`.
+"""
+function _complementary_type(system_type::Type{<:AbstractSystem})
+    type_string = string(Base.typename(system_type))
```
blegat: I guess Julia will have troubles inferring the result and it might have inference issue. Maybe a generated function would solve it ?

blegat: You can test it in the tests with `@inferred`

ueliwechsler:
Thanks for the review! I will address the comments shortly.

src/discretize.jl
ueliwechsler: It doesn't look like that my implementation is type stable
 ```julia
@inferred MathematicalSystems._complementary_type(LinearDiscreteSystem)
ERROR: return type Type{LinearContinuousSystem} does not match inferred return type Any
```

I think it works with a generated function
````julia
@generated function _complementary_type(system_type::Type{<:AbstractSystem})
    type_string = string(Base.typename(system_type))
    if supertype(system_type) == AbstractDiscreteSystem
        type_string =  replace(type_string, "Discrete"=>"Continuous")
    else
        type_string =  replace(type_string, "Continuous"=>"Discrete")
    end
    return Meta.parse(type_string)
end

>julia @inferred _complementary_type(LinearDiscreteSystem) #true

File: src/discretize.jl (hunk 1 above)
Comment (ueliwechsler) [reply src/discretize.jl]: I am not sure if a understand correctly. You think we can get rid of the other `_discretize` methods, or the `spzeros`'s in the other `_discretize` methods.

File: test/discretize.jl (hunk 5 above)
Comment (ueliwechsler) [reply test/discretize.jl]: It has to do with #124.
But you are right, it's a useless comment

File: test/discretize.jl (hunk 6 above)
Comment (ueliwechsler) [reply test/discretize.jl]: Maybe, we could add them in the traits definition by looping over something similar to `[Z, Type{Z}]`.

File: src/discretize.jl (hunk 1 above)
Comment (schillic) [reply src/discretize.jl]: I meant:
```julia
function _discretize(ΔT::Real,
                     A::AbstractMatrix;
                     B::AbstractMatrix=spzeros(n, n),
                     c::AbstractVector=spzeros(n),
                     D::AbstractMatrix=spzeros(n, n),
                     algorithm=:exact)
    ...
    return [A_d, B_d, c_d, D_d]
end
```
But now I realized that your result depends on the method, so that does not work. You could instead try

```julia
function _discretize(ΔT::Real,
                     A::AbstractMatrix;
                     B=nothing,
                     c=nothing,
                     D=nothing,
                     algorithm=:exact)
    if algorithm == :exact
        A_d = exp(A*ΔT)
        indices = [1]
        if B != nothing
            B_d = inv(A)*(A_d - I)*B
            push!(indices, 2)
        else
            B_d = nothing
        end
        if c != nothing
            c_d = inv(A)*(A_d - I)*c
            push!(indices, 3)
        else
            c_d = nothing
        end
        if D != nothing
            D_d = inv(A)*(A_d - I)*D
            push!(indices, 4)
        else
            D_d = nothing
        end
    else
        ...
    end
    return [A_d, B_d, c_d, D_d][indices]
end
```

but I agree that this is not optimal and you probably cannot use the keyword arguments from your other code anyway.

src/discretize.jl
ueliwechsler: I tried to fix this problem and added a `@inferred` in the tests. I had to add the method `typename` to use the method as `_complementary_type(typename(system))` for a variable `system` which is the object and not the the type, i.e.`system= LinearContinuousSystem(A)`

src/discretize.jl
```diff
@@ -0,0 +1,190 @@
+using LinearAlgebra: inv, rank
+using SparseArrays: spzeros
+
+"""
+    typename(system::AbstractSystem)
+
+    Returns base-type of `system` without information parameter information.
```
schillic:
```suggestion
Returns the base type of `system` without parameter information.
```
Again the indentation is wrong here.

src/discretize.jl
```diff
@@ -0,0 +1,173 @@
...
+        else
+          # A is not invertible, use approximative discretizaion
+          algorithm = :euler
+        end
+    end
+    disc_nonset_values = _discretize(cont_nonset_values..., ΔT; algorithm=algorithm)
+    set_values = [getfield(sys, f) for f in filter(!noset, fields)]
+    discrete_type = _complementary_type(sys)
+    return discrete_type(disc_nonset_values..., set_values...)
+end
```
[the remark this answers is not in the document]
ueliwechsler: I added some.

src/discretize.jl
ueliwechsler: 🙈

test/discretize.jl
ueliwechsler: @schillic  should I change something here or add an issue for the trait?

src/discretize.jl
```diff
@@ -0,0 +1,173 @@
...
+    If A is invertible:
+    The exact discretization is calculated by solving the integral for
+    `t = [t, t+ΔT]` for a fixed input `u` and fixed noise realisation `w` which
+    writes as `x⁺ = Aᵈx + Bᵈu + cᵈ + Dᵈu` where `Aᵈ = exp(A⋅ΔT)`,
+    `Bᵈ = inv(A)⋅(Aᵈ - I)⋅B`, `cᵈ = inv(A)⋅(Aᵈ - I)⋅c` and `Dᵈ = inv(A)⋅(Aᵈ - I)⋅D`.
+
+    If A is not invertible:
+    A first order approximation of the exact discretiziation, the euler
+    discretization can be apllied which writes as `x⁺ = Aᵈx + Bᵈu + cᵈ + Dᵈu`
+     where  `Aᵈ = I + ΔT⋅A`, `Bᵈ = ΔT⋅B`, `cᵈ = ΔT⋅c` and `Dᵈ = ΔT⋅D`.
```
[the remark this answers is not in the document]
schillic: As far as I know, you should be able to use `Δ` etc. in math mode, too.

test/discretize.jl
schillic: This was just a comment. I created #130 and if somebody finds it important enough to add those methods, these things can be replaced. But you do not need to do it here.

src/discretize.jl
ueliwechsler: Exactly, thanks for the example!
But I doesn't nicely work with the existing structure, which is why I prefer the other approach. But I am open to other suggestions.

schillic: I agree to keep your version.

ueliwechsler:
Is there anything I have to add or change? Apart from the docstring for the `_discretize` method?

src/MathematicalSystems.jl
```diff
@@ -158,4 +158,11 @@ export successor
 
 include("successor.jl")
 
+#===================================
+Discretization for affine systems
+====================================#
+export discretize, typename
```
schillic: I would not export `typename` since it is not specific to this library.

src/discretize.jl
```diff
@@ -0,0 +1,192 @@
...
+
+- `system_type` -- type of `AbstractSystem`
+
+### Ouput
+
+Return complementary type of `system_type`.
+
+### Note
+To get the `_complementary_type` of a `system<:AbstractSystem` use
+`_complementary_type(typename(system))`.
```
schillic: This docstring does not explain what "complementary" means.

src/discretize.jl
```diff
@@ -0,0 +1,192 @@
...
+"""
+@generated function _complementary_type(type::Type{<:AbstractSystem})
+    # type is a Type{<:AbstractSystem}, e.g. Type{AffineContinuousSystem}
```
schillic:
```suggestion
```
I think this should be clear.

src/discretize.jl
```diff
@@ -0,0 +1,192 @@
...
+    # extract the system type information with
```
schillic:
```suggestion
```
This comment is not helpful to me.

src/discretize.jl
```diff
@@ -0,0 +1,192 @@
...
+"""
+     discretize(system::AbstractContinuousSystem, ΔT::Real; algorithm=:default)
+
+Discretization of a `isaffine` `AbstractContinuousSystem` to a
+`AbstractDiscreteSystem` with discretization time `ΔT` using the exact
+discretization algorithm if possible.
+
+### Input
+
+- `system` -- a affine continuous system
```
schillic:
```suggestion
- `system` -- an affine continuous system
```

src/discretize.jl
```diff
@@ -0,0 +1,192 @@
...
+- `ΔT` -- discretization time
+- `algorithm` -- (optional, default=`:default`) discretization algorithm
```
schillic:
```suggestion
- `algorithm` -- (optional, default: `:default`) discretization algorithm
```

src/discretize.jl
```diff
@@ -0,0 +1,192 @@
...
+Returns a discretization of the input system `system` with discretization time `ΔT`.
+
+### Algorithm
+
+Consider a `NoisyAffineControlledContinuousSystem` with system dynamics
+``x' = Ax + Bu + c + Du``.
+
+If A is invertible:
+The exact discretization is calculated by solving the integral for
+``t = [t, t + \\Delta T]`` for a fixed input `u` and fixed noise realisation `w`
```
schillic: I would replace those `\\Delta`s by `Δ`.

src/discretize.jl
```diff
@@ -0,0 +1,192 @@
...
+discretization, can be applied, which writes as ``x^+ = A^d x + B^d u + c^d + D^d u``
+where  ``A^d = I + \\Delta T \\cdot A``, ``B^d = \\Delta T \\cdot B``,
+``c^d = ΔT \\cdot c`` and ``D^d = \\Delta T \\cdot D``.
+"""
+function discretize(system::AbstractContinuousSystem, ΔT::Real; algorithm=:default)
+    noset(x) = x ∉ [:X,:U,:W]
+    # get all fields from system
+    fields = collect(fieldnames(typeof(system)))
+    # get fields of system that are parameter of the system dynamics (no sets)
+    # i.e., all fields that needs to be discretized
```
schillic:
```suggestion
    # i.e., all fields that need to be discretized
```

src/discretize.jl
```diff
@@ -0,0 +1,192 @@
...
+
+
+If A is not invertible:
+A first order approximation of the exact discretiziation, the euler
+discretization, can be applied, which writes as ``x^+ = A^d x + B^d u + c^d + D^d u``
+where  ``A^d = I + \\Delta T \\cdot A``, ``B^d = \\Delta T \\cdot B``,
+``c^d = ΔT \\cdot c`` and ``D^d = \\Delta T \\cdot D``.
+"""
+function discretize(system::AbstractContinuousSystem, ΔT::Real; algorithm=:default)
+    noset(x) = x ∉ [:X,:U,:W]
```
schillic: Instead of describing what it is *not*, I would negate the predicate and change the name to the positive descripton `matrices` (and rephrase the comments) because the non-matrices are not automatically sets (consider the constant term of an affine system and who knows what other fields the systems will have in the future).

src/discretize.jl
```diff
@@ -0,0 +1,192 @@
...
+``B^d = A^{-1}(A^d - I)B``,
+``c^d = A^{-1}(A^d - I)c`` and
+``D^d = A^{-1}(A^d - I)D``.
+
+
+If A is not invertible:
+A first order approximation of the exact discretiziation, the euler
+discretization, can be applied, which writes as ``x^+ = A^d x + B^d u + c^d + D^d u``
+where  ``A^d = I + \\Delta T \\cdot A``, ``B^d = \\Delta T \\cdot B``,
+``c^d = ΔT \\cdot c`` and ``D^d = \\Delta T \\cdot D``.
```
schillic: Just to make clear that this is nothing fancy, can you add pointers to sources where this is described? Wikipedia is fine.

src/discretize.jl
```diff
@@ -0,0 +1,327 @@
...
+- `algorithm` -- (optional, default: `:default`) discretization algorithm
+
+### Output
+
+Returns a discretization of the input system `system` with sampling time `ΔT`.
+
+### Algorithm
+
+Consider a `NoisyAffineControlledContinuousSystem` with system dynamics
+``x' = Ax + Bu + c + Du``.
```
blegat: `Du` -> `Dw`

src/discretize.jl
```diff
@@ -0,0 +1,327 @@
...
+
+If A is invertible:
+The exact discretization is calculated by solving the integral for
+``t = [t, t + ΔT]`` for a fixed input `u` and fixed noise realisation `w`.
+The resulting discretization writes as
+``x^+ = A^d x + B^d u + c^d  + D^d u`` where
```
blegat: D^d u -> D^d w

src/discretize.jl
```diff
@@ -0,0 +1,327 @@
...
+``A^d = \\exp^{A \\cdot ΔT}``,
+``B^d = A^{-1}(A^d - I)B``,
+``c^d = A^{-1}(A^d - I)c`` and
+``D^d = A^{-1}(A^d - I)D``.
+
+If A is not invertible:
+A first order approximation of the exact discretiziation, the euler
+discretization, can be applied, which writes as ``x^+ = A^d x + B^d u + c^d + D^d u``
```
blegat: u -> w

blegat:
Switching from exact to order one approximation based on whether `A` is singular seems unexpected from the user viewpoint. I would rather have the algorithms given as a third positional argument. The algorithm can be a subtype of `AbstractDiscretizationAlgorithm` so that additional algorithms can be added outside of this package and implemented through multiple dispatch.
If the algorithm is `ExactDiscretization()` and the matrix is nonsingular, we can still compute the exact discretization, e.g. https://math.stackexchange.com/questions/658276/integral-of-matrix-exponential

ueliwechsler:
> so that additional algorithms can be added outside of this package and implemented through multiple dispatch

Good idea. I will work on it.  What is your opinion on adding `ΔT` as a field of the concrete Discritziation type, i.e. `ExactDiscretization(ΔT)`?

> If the algorithm is `ExactDiscretization()` and the matrix is nonsingular, we can still compute the exact discretization, e.g. https://math.stackexchange.com/questions/658276/integral-of-matrix-exponential

Thanks for the link. However, this computation seems quite involved (see e.g. https://discourse.julialang.org/t/jordan-form-of-a-matrix/7123/6), therefore, I will throw an error if the state matrix is singular and open an issue.

blegat:
I prefer leaving the ΔT parameter separate from the algo so that the user can easily change it in a code that is generic on the algorithm.

> I will throw an error if the state matrix is singular and open an issue.

Sounds good, we can leave that as future work

ueliwechsler:
I try to rebase master, in order to merge the PR, okay?

mforets:
Sorry for my late reply, i've been taking it easy as i'm on holidays. I'd like to have a close look before you merge, i will do it asap. I don't think it's necessary that you rebase..

schillic:
> I'd like to have a close look before you merge, i will do it asap.

Same, but I might not get there before the weekend. There are mainly some typos in the new parts that I would like to fix.

> I don't think it's necessary that you rebase

Agreed, I do not see any external changes here.

ueliwechsler:
Sorry, I did not wanna stress you. No hurry, enjoy your holidays Marcelo :)

src/discretize.jl
```diff
@@ -0,0 +1,367 @@
...
+
+Returns the base type of `system` without parameter information.
+
+### Input
+
+- `system` -- `AbstractSystem`
+
+### Output
+
+Returns base-type of `system`.
```
schillic:
```suggestion
The base type of `system`.
```

src/discretize.jl
```diff
@@ -0,0 +1,367 @@
...
+function typename(system::AbstractSystem)
+    return Base.typename(typeof(system)).wrapper
+end
+
+"""
+    _complementary_type(system_type::Type{<:AbstractSystem})
+
+Return the complementary type of a system type `system_type`.
+
+There are two main subclasses of abstract types continuous types and discrete
```
schillic:
```suggestion
There are two main subclasses of abstract types: continuous types and discrete
```

src/discretize.jl
```diff
@@ -0,0 +1,367 @@
...
+### Input
+
+- `system_type` -- type of `AbstractSystem`
+
+### Ouput
+
+Return complementary type of `system_type`.
+
+### Note
+To get the `_complementary_type` of a `system<:AbstractSystem` use
```
schillic:
```suggestion

To get the `_complementary_type` of a `system::AbstractSystem` use
```

src/discretize.jl
```diff
@@ -0,0 +1,367 @@
...
+
+Exact discretization algorithm for affine systems.
+
+### Algorithm
+
+Consider a `NoisyAffineControlledContinuousSystem` with system dynamics
+``x' = Ax + Bu + c + Dw``.
+
+The exact discretization is calculated by solving the integral for
+``t = [t, t + ΔT]`` for a fixed input `u` and fixed noise realisation `w`.
```
schillic:
```suggestion
``t = [t, t + ΔT]`` for a fixed input `u` and fixed noise realization `w`.
```

src/discretize.jl
```diff
@@ -0,0 +1,367 @@
...
+The resulting discretization writes as
+``x^+ = A^d x + B^d u + c^d  + D^d w`` where
+``A^d = \\exp^{A \\cdot ΔT}``,
+``B^d = A^{-1}(A^d - I)B``,
+``c^d = A^{-1}(A^d - I)c`` and
+``D^d = A^{-1}(A^d - I)D``.
+
+The algorithm described above is a well known results from the literature.
```
schillic:
```suggestion
The algorithm described above is a well known result from the literature [1].
```

src/discretize.jl
```diff
@@ -0,0 +1,367 @@
...
+Consider [1] as a source for further information.
```
schillic:
```suggestion
```

src/discretize.jl
```diff
@@ -0,0 +1,367 @@
...
+Consider a `NoisyAffineControlledContinuousSystem` with system dynamics
+``x' = Ax + Bu + c + Dw``.
+
+The euler discretization, a first order approximation of the exact
+discretiziation [`ExactDiscretization`](@ref), writes as
+``x^+ = A^d x + B^d u + c^d + D^d w``
+where  ``A^d = I + ΔT \\cdot A``, ``B^d = ΔT \\cdot B``,
+``c^d = ΔT \\cdot c`` and ``D^d = ΔT \\cdot D``.
+
+The algorithm described above is a well known results from the literature.
```
schillic:
```suggestion
The algorithm described above is a well known result from the literature [1].
```

src/discretize.jl
```diff
@@ -0,0 +1,367 @@
...
+Consider [1] as a source for further information.
```
schillic:
```suggestion
```

src/discretize.jl
```diff
@@ -0,0 +1,367 @@
...
+
+[1] https://en.wikipedia.org/wiki/Discretization#Approximations
+"""
+struct EulerDiscretization <: AbstractDiscretizationAlgorithm end
+
+"""
+    discretize(system::AbstractContinuousSystem, ΔT::Real,
+               algorithm::AbstractDiscretizationAlgorithm=ExactDiscretization())
+
+Discretization of a `isaffine` `AbstractContinuousSystem` to a
```
schillic: Should we check `isaffine(system)`?

src/discretize.jl
```diff
@@ -0,0 +1,367 @@
...
+"""
+function discretize(system::AbstractContinuousSystem, ΔT::Real,
+                    algorithm::AbstractDiscretizationAlgorithm=ExactDiscretization())
+    sets(x) = x ∈ [:X,:U,:W]
+    matrices(x) = x ∈ [:A,:B,:b,:c,:D]
+
+    # get all fields from system
+    fields = collect(fieldnames(typeof(system)))
+
+    # get the fields of `system` that are parameter of the affine system dynamics
```
schillic:
```suggestion
    # get the fields of `system` that are parameters of the affine system dynamics
```

src/discretize.jl
```diff
@@ -0,0 +1,367 @@
...
+    values_disc = _discretize(algorithm, ΔT, values_cont...)
+
+    # get the fields of `system` that are sets
+    set_values = [getfield(system, f) for f in filter(sets, fields)]
+
+    # get corresponding discrete type of `system`
+    discrete_type = _complementary_type(typename(system))
+
+    # build the new discrete type with the discretized and set values
+    return discrete_type(values_disc..., set_values...)
```
schillic: This relies on the constructors taking the arguments in that order and that there are no further arguments, right? I fear that we may break this invariant during a future refactoring. Also, this prevents you from using this function with your own system type that has additional fields.

#### Suggestion
Make `discrete_type` an input parameter of the function, with the default being the implementation you have here.
```julia
function discretize(system::AbstractContinuousSystem, ΔT::Real,
           algorithm::AbstractDiscretizationAlgorithm=ExactDiscretization(),
           constructor=_default_complementary_constructor(system))
    ...
    return constructor(values_disc, set_values)
end

# returns a function with signature 'Vector × Vector -> AbstractDiscreteSystem'
function _default_complementary_constructor(system)
    discrete_type = _complementary_type(typename(system))
    return (values_disc, set_values) -> discrete_type(values_disc..., set_values...)
end
```

src/discretize.jl
```diff
@@ -0,0 +1,367 @@
...
+end
+
+"""
+    _discretize(::AbstractDiscretizationAlgorithm, ΔT::Real
+                A::AbstractMatrix, B::AbstractMatrix, c::AbstractVector, D::AbstractMatrix)
+
+Implementation of the discretization algorithm defined by the first input argument
+with sampling time `ΔT`.
```
schillic: This docstring is associated with the `::ExactDiscretization` method. I think you need to write the method stub with `::AbstractDiscretizationAlgorithm` and move the docstring there (and just keep the two concrete methods without any docstring).

src/discretize.jl
```diff
@@ -0,0 +1,367 @@
...
+        A_d = exp(A*ΔT)
+        Matr = inv(A)*(A_d - I)
+        B_d = Matr*B
+        c_d = Matr*c
+        D_d = Matr*D
+    else
+        error("exact discretization for singular state matrix, i.e. A is non-invertible,"*
+              " not implemented yet, please use algorithm `EulerDiscretization`")
+    end
+    return [A_d, B_d, c_d, D_d]
```
schillic: Do you need to create a `Vector` as return value in the `_discretize` methods? Just removing the brackets might be better (but I do not know).

src/discretize.jl
ueliwechsler: What do you mean by 

> think you need to write the method stub with `::AbstractDiscretizationAlgorithm` and move the docstring there

I don't quite get it.

schillic:
```julia
"""
    old docstring
"""
function _discretize(::AbstractDiscretizationAlgorithm, ΔT::Real,
                     A::AbstractMatrix, B::AbstractMatrix, c::AbstractVector,
                     D::AbstractMatrix) end  # note the `end`

function _discretize(::ExactDiscretization, ΔT::Real, A::AbstractMatrix,
                     B::AbstractMatrix, c::AbstractVector, D::AbstractMatrix)
    # old code as before
end

File: src/discretize.jl (hunk 33 above)
Comment (ueliwechsler) [reply src/discretize.jl]: Yes, this makes sense.

File: src/discretize.jl (hunk 35 above)
Comment (ueliwechsler) [reply src/discretize.jl]: I like your suggestion. I will add it.

File: src/discretize.jl (hunk 36 above)
Comment (ueliwechsler) [reply src/discretize.jl]: Got it, thanks.

File: src/discretize.jl (hunk 37 above)
Comment (ueliwechsler) [reply src/discretize.jl]: It should work without the `[]`, I will test and remove it if possible.

Comment (schillic) [reply src/discretize.jl]: There are more occurrences of `[...]`. Are they needed?

Comment (ueliwechsler) [reply src/discretize.jl]: just removing the `[...]` does not work, but I should make it consistent overall `_discretize` calls. I will have another look at it.

Comment (ueliwechsler) [reply src/discretize.jl]: I added the `[...]` back in. The `discretize(disc_values..., set_values...)` for the case of a single Matrix does not properly work otherwise.

Comment (schillic) [reply src/discretize.jl]: Alright, thanks for trying, though :+1:

File: src/discretize.jl (hunk 8 above)
Comment (ueliwechsler) [reply src/discretize.jl]: @blegat Do you agree with my changes?

File: src/discretize.jl [hunk 38]
```diff
@@ -0,0 +1,383 @@
...
+``x' = Ax + Bu + c + Dw``.
+
+The exact discretization is calculated by solving the integral for
+``t = [t, t + ΔT]`` for a fixed input `u` and fixed noise realization `w`.
+The resulting discretization writes as
+``x^+ = A^d x + B^d u + c^d  + D^d w`` where
+``A^d = \\exp^{A \\cdot ΔT}``,
+``B^d = A^{-1}(A^d - I)B``,
+``c^d = A^{-1}(A^d - I)c`` and
+``D^d = A^{-1}(A^d - I)D``.
```
mforets: It should be stated here that ``A`` is assumed to be invertible.

src/discretize.jl
```diff
@@ -0,0 +1,383 @@
...
+abstract type AbstractDiscretizationAlgorithm end
+
+"""
+    ExactDiscretization <: AbstractDiscretizationAlgorithm
+
+Exact discretization algorithm for affine systems.
+
+### Algorithm
+
+Consider a `NoisyAffineControlledContinuousSystem` with system dynamics
```
mforets: The system `NoisyAffineControlledContinuousSystem` is chosen because it comprises the other systems relevant to this function as special cases. However, that didn't seem to me clear at once. We can state this, eg.

> This algorithm consists of integrating the continuous differential equation over a specified time interval to obtain an associated discrete differential equation. The algorithm applies to any system of the form ``x' = Ax + Bu + c + Dw``, and other system types e.g. linear systems, ``x' = Ax``, are also included as particular cases. Without loss of generality, consider a `NoisyAffineControlledContinuousSystem` ....

src/discretize.jl
```diff
@@ -0,0 +1,383 @@
...
+``x' = Ax + Bu + c + Dw``.
+
+The exact discretization is calculated by solving the integral for
+``t = [t, t + ΔT]`` for a fixed input `u` and fixed noise realization `w`.
```
mforets: ... by integrating on both sides of the continuous ODE over the time span ``[t, t + ΔT]``, for a fixed input `u` and fixed noise realization `w` at time `t`.

src/discretize.jl
```diff
@@ -0,0 +1,383 @@
...
+struct ExactDiscretization <: AbstractDiscretizationAlgorithm end
+
+"""
+    EulerDiscretization <: AbstractDiscretizationAlgorithm
+
+Euler discretization algorithm for affine systems.
+
+### Algorithm
+
+Consider a `NoisyAffineControlledContinuousSystem` with system dynamics
```
mforets: The same minor comment applies as with `ExactDiscretization` regarding an introductory sentence for this section.

src/discretize.jl
```diff
@@ -0,0 +1,383 @@
...
+``x' = Ax + Bu + c + Dw``.
+
+The euler discretization, a first order approximation of the exact
```
mforets:
```suggestion
The Euler discretization, a first order approximation of the exact
```

src/discretize.jl
```diff
@@ -0,0 +1,383 @@
...
+discretiziation [`ExactDiscretization`](@ref), writes as
```
mforets:
```suggestion
discretization [`ExactDiscretization`](@ref), writes as
```

src/discretize.jl
```diff
@@ -0,0 +1,383 @@
...
+``x^+ = A^d x + B^d u + c^d + D^d w``
+where  ``A^d = I + ΔT \\cdot A``, ``B^d = ΔT \\cdot B``,
```
mforets: (it's a minor thing but i wouldn't use the `\\cdot` in these products, as that symbol is more often reserved to the inner product -- cf. the wikipedia article; you can use one or more tilde ``~`` to leave an extra small space if the default one is not enough)

mforets:
About the code for `discretize` taken in this PR, while the implementation is clever as it's general and works, I would advise against this approach and instead use dispatch with the system types. The final LOC are more, for sure, but that's a reasonable price to pay. There are two things:

- performance: specialized code for systems will be more efficient, e.g. one doesn't have to move the vector `[A_d, B_d, c_d, D_d]` around and allocate extra arrays when it's not necessary, 

- extendability: suppose that someone from an external library defines a new kind of system for which discretization makes sense. However, one is imposing strong constraints on the fields names (`:A`,`:B`,`:b`,`:c`,`:D`) to make it applicable in general (i refer to lines [151 and 152](https://github.com/JuliaReach/MathematicalSystems.jl/pull/122/files#diff-38e90d26af8d0722d139aa4d6bdb1d64R151) in `discretize`).

mforets:
Thanks for the hard work, merging!

I'll open separate PRs this morning for some comments left.

PR #124, opened 2019-12-22
Make fieldname of AffineSystems and ConstrainedAffineSystems more consistent

ueliwechsler:
This PR renames the field `b` in `AffineSystems` and `ConstrainedAffineSystems` to `c`.

It makes more sense to name the constant term in `AffineSystems` `c` for all systems, independent if they are controlled or not. Especially for the soon to come `@system` macro, it is convenient if the same letters have the same meaning for all types.


schillic:
I am not sure this breaking change is necessary. I would rather add getter functions. But I do not have a strong opinion here.

mforets:
> Especially for the soon to come @system macro, it is convenient if the same letters have the same meaning for all types.

True, but i think it's preferable if the implementation does not rely on the field name, especially if it is intended to work for different system's types. Adding a getter function `affine_term` can do the job in a more robust way, as suggested by @schillic in the comment above.

blegat:
I would write the convention somewhere in the doc or README, IIUC, it is
* `A` for the state
* `B` for the control
* `c` for the constant term
* `D` for the noise
We also need a convincing argument on why it's not
* `d` for the constant term
* `C` for the noise

ueliwechsler:
> Adding a getter function `affine_term` can do the job in a more robust way, as suggested by @schillic in the comment above.

To make the implementation of `@system` generic, we would need a method `affine_field_value` that returns `:c` and the same for all other fields. So I would prefer to have a naming convention.

We could still have add the methods `state_matrix`, `input_matrix`, `noise_matrix` and `affine_term`.

> * We also need a convincing argument on why it's not

Is there a consensus on how to name an affine system in general? 
I feel pretty good about `A` and `B` but concerning  `c` and `D` or `C` and `d`, these letters are normally used in the output map `y = Cx + Du`.

schillic:
> True, but i think it's preferable if the implementation does not rely on the field name, especially if it is intended to work for different system's types.

I am not worried that the implementation relies on the field names. A package may make assumptions about the package's own code. It is of course easier for refactorings if you do not, but this is only an internal issue.

My point was that user code will break. Currently we do not have any getter functions, and so the field names are our only API. Consequently, *any* implementation that makes reasonable use of the systems must access the field names and thus must inherently break.

>  We also need a convincing argument on why it's not `d` for the constant term and `C` for the noise

Exactly, we may see reason to rename them again at some point. A stable API would make such changes invisible to the user.

mforets:
This library is still young and adding getter functions such as the ones proposed in this thread is always a good idea. I've opened https://github.com/JuliaReach/MathematicalSystems.jl/issues/126 for this purpose.

I'd like to merge this PR to incorporate the changes as part of the naming convention.

schillic:
> I'd like to merge this PR

okay

mforets:
Merging.. I plan to do #126 and #127 asap, of course.

PR #125, opened 2019-12-24
Add @system macro

### LINKED ISSUE #76: Add @system macro
Initially i thought about: `@system x -> A*x` to define a `LinearContinuousSystem`. This is saying two things:`@system` defines a continuous ODE, and the anonymous function `x -> f(x, u)` refers to the right-hand side map of the ODE.

Alternatively let `@continous_system x -> A*x` and `@discrete_system x-> A*x` but it is kind of verbose and it's what we wanted to avoid with the macros.. or `@cont_sys` and `@cont_sys`..

Or a specific convention could be used for discrete systems, such as `@system x_k -> A*x_k`.

ueliwechsler:
Closes #76.


schillic:
Does this close #76?

ueliwechsler:
Yes, once it is done.

src/macros.jl
```diff
@@ -89,3 +89,308 @@ macro map(ex)
 
     throw(ArgumentError("unable to match the given expression to a known map type"))
 end
+
+
+
+# @System Macro ================================================================
+# Autors: @mforets, @ueliwechsler
```
mforets: I'll remove this as we don't usually write authors in the code. This information is available on git.

src/macros.jl
```diff
@@ -89,3 +89,308 @@ macro map(ex)
...
+using LinearAlgebra: I
+using MacroTools: @capture
```
mforets: Using `MacroTools.@capture` is an alternative to the `Espresso` `matchex` stuff. I'm considering using `@capture` in the `@map` macro too. It simplifies things a bit.

src/macros.jl
```diff
@@ -89,3 +89,308 @@ macro map(ex)
 
     throw(ArgumentError("unable to match the given expression to a known map type"))
 end
+
+
+
+# @System Macro ================================================================
+# Autors: @mforets, @ueliwechsler
+using LinearAlgebra: I
```
mforets: We can also move the `using` and `export` statements on top of the file as usual.

src/macros.jl
```diff
@@ -89,3 +89,308 @@ macro map(ex)
...
+using MacroTools: @capture
+using InteractiveUtils: subtypes
+export @system
+
+function corresponding_type(abstract_type, fields::Tuple)
+    TYPES = subtypes(abstract_type)
+    TYPES_FIELDS = fieldnames.(TYPES)
+    is_in(x, y) = all([el ∈ y for el in x])
```
mforets: Allocating the array is not needed using the two-argument form of all:
```suggestion
    is_in(x, y) = all(el -> el ∈ y, x)
```
But more simply put, i think we can use `is_in(x, y) = x ⊆ y` (see the documentation of `Base.issubset`). So we can remove `is_in(x, y)` altogether and write `x ⊆ fields` and  `fields ⊆ x` below.

src/macros.jl
```diff
@@ -89,3 +89,308 @@ macro map(ex)
...
+    idx = findall(x -> is_in(x, fields) && is_in(fields,x), TYPES_FIELDS)
+    if length(idx) == 0
+        error("The entry $(fields) does not match a MathematicalSystem structure.")
```
mforets: Here i would throw an `ArgumentError` with the message, as in `throw(ArgumentError("...."))`.

src/macros.jl
```diff
@@ -89,3 +89,308 @@ macro map(ex)
...
+
+# @System Macro ================================================================
+# Autors: @mforets, @ueliwechsler
+using LinearAlgebra: I
+using MacroTools: @capture
+using InteractiveUtils: subtypes
+export @system
+
+function corresponding_type(abstract_type, fields::Tuple)
+    TYPES = subtypes(abstract_type)
```
mforets: This information could be stored in a constant table (and evaluated only once).

src/macros.jl
```diff
@@ -89,3 +89,308 @@ macro map(ex)
...
+
+
+# @System Macro ================================================================
+# Autors: @mforets, @ueliwechsler
+using LinearAlgebra: I
+using MacroTools: @capture
+using InteractiveUtils: subtypes
+export @system
+
+function corresponding_type(abstract_type, fields::Tuple)
```
mforets: The general idea of `corresponding_type` is that you throw it a tuple of fields and it finds a match with the fields of some system. While the idea is nice and works automatically, it needs that `fields` have been normalized to match the naming convention (correct me if i'm wrong).

I've thought of an alternative approach by pattern matching with `@capture` in a more robust way, without having to preparse the fields. We do need to do some work on the system's side defining the templates, it won't magically work. For instance, a linear continuous system needs to define its template form, `:(x' = A_*x)` so we can pattern-match with that expression.

src/macros.jl
```diff
@@ -89,3 +89,308 @@ macro map(ex)
...
+    return dims
+end
+
+# return `true` if the given expression `expr` corresponds to an equation `lhs = rhs`
+# and `false` otherwise. This function just detects the presence of the symbol `=`.
+function is_equation(expr)
+    return @capture(expr, lhs_ = rhs_)
+end
+
+# return the tuple `(true, AT)` if the given expression `expr` corresponds to a dynamic equation,
```
mforets: this part of the docstring is outdated

src/macros.jl
mforets: .. or we use `Base.issetequal` :smile:

src/macros.jl
ueliwechsler: In the way it is implemented at the moment, you are right. 
But if there is a functionality like `state_matrix_field(::AbstractSystem) = :A` this could be made more robust and indepentend of the `fields`.

> I've thought of an alternative approach

I tried to make it work like this in the beginning, but then I fought it to be too tedious to implement and not automatically working, therefore, I change to the current approach.
You have to be careful, that the approach can also handle `:(x⁺ = A*x)` since this is a significantly different expression than  `:(x' = A*x)` because `'` is a "special operator". But I am open for more elegant and robust approaches. 
However, for this PR, I would stick to the original idea and add these changes in subsequent PRs.

src/macros.jl
```diff
@@ -89,3 +98,501 @@ macro map(ex)
...
+end
+
+"""
+    extract_sum(summands, state::Symbol, noise::Symbol)
+
+Given a array of expression `summands` which consists of one or more elements
+which either are mutliplication or a symbol, the corresponding fields of the
+affine system and the variable use are extracted.
+The state variable is parsed as `state`, the noise variable as `noise` and the input
+variable as everything else.
```
schillic: I find it dangerous to parse inputs as "everything else".

src/macros.jl
```diff
@@ -89,3 +98,501 @@ macro map(ex)
...
+
+julia> extract_sum(:(A1*x7+ B1*w + B2*w8), :x7, :w8)
+[(:A1, :A), (:B1, :B), (:B2, :D)]
+```
+"""
+function extract_sum(summands, state::Symbol, noise::Symbol)
+    params = Any[]
+    for summand in summands
+        if @capture(summand, array_ * var_)
+            @show array, var
```
schillic: reminder to eventually remove the `@show`s

src/macros.jl
```diff
@@ -89,3 +98,501 @@ macro map(ex)
...
+A system that best matches the given expressions.
+
+### Note
+
+The `expr` consist of one or several of the following elements:
+- continuous dynamic equation: `x' = Ax `
+- discrete dynamic equation: `x⁺ = Ax `
+- sets: `x ∈ X`
+- dimensionality: `dim: 1` or `dim = 1`
+- defining the noise variable: `noise: w`, `noise= w`
```
```suggestion
- defining the noise variable: `noise: w`, `noise = w`
```

src/macros.jl
```diff
@@ -89,3 +98,501 @@ macro map(ex)
...
+
+The dynamic equation is parsed as following. The variable on the left hand side
+corresponds to the state variable, the noise variable is by default `w`, if not
+specified differently, and the input variable is the arbitrary remaining variable.
+If we want to change the default of the noise variable, this can be done by adding
```
```suggestion
If we want to change the default name of the noise variable, this can be done by adding
```

src/macros.jl
```diff
@@ -89,3 +98,501 @@ macro map(ex)
...
+the term `noise: var` where `var` corresponds to the new value for the noise variable.
```
```suggestion
the term `noise: var` where `var` corresponds to the new name of the noise variable.
```

src/macros.jl
```diff
@@ -89,3 +98,501 @@ macro map(ex)
...
+
+If asterisk are used to separated the terms of the equation, the input variable
```
```suggestion
If asterisks are used to separate the terms of the equation, the input variable
```

src/macros.jl
```diff
@@ -89,3 +98,501 @@ macro map(ex)
...
+can be any valid variable, i.e. `x⁺ = A*x + B*u_extra_long` if no asaterisk are
```
```suggestion
can be any valid variable, i.e. `x⁺ = A*x + B*u_extra_long`. If no asterisks are
```

src/macros.jl
```diff
@@ -89,3 +98,501 @@ macro map(ex)
...
+
+### Examples
+
+Let us first create a continuous linear system using this macro:
+
+```jldoctest
+julia> A = [1. 0; 0 1.]
+julia> @system(x' = A*x)
+LinearContinuousSystem{Array{Int64,2}}([1. 0; 0 1.])
+```
```
schillic: You need to use `;` for lines where you do not want to show outputs. Still, you need to add an empty line afterward. And I guess that `[1. 0; 0 1.]` is different from the actual Julia output.

src/macros.jl
```diff
@@ -89,3 +98,501 @@ macro map(ex)
...
+a discrete system can be defined by using  `⁺`
```
```suggestion
A discrete system can be defined by using  `⁺`:
```

src/macros.jl
```diff
@@ -89,3 +98,501 @@ macro map(ex)
...
+
+```jldoctest
+julia> A = [1. 0; 0 1.]
+julia> @system(x⁺ = A*x)
+LinearDiscreteSystem{Array{Int64,2}}([1. 0; 0 1.])
+```
+
+Additionally, a set definition `x ∈ X` can be added to create a constrained system
```
```suggestion
Additionally, a set definition `x ∈ X` can be added to create a constrained system.
```

src/macros.jl
```diff
@@ -89,3 +98,501 @@ macro map(ex)
...
+For example, an  controlled affine system with state and input constrained writes
```
```suggestion
For example, a discrete controlled affine system with constrained states and inputs writes
```

src/macros.jl
```diff
@@ -89,3 +98,501 @@ macro map(ex)
...
+julia> A = [1. 0; 0 1.]
+julia> B = Matrix([1. 0.5]')
+julia> c = [1., 1.5]
+julia> X = BallInf(zeros(2), 10.)
+julia> U = BallInf(zeros(1), 2.)
+julia> @system(x⁺ = A*x + B*u + c, x∈X, u∈U)
+```
+
+For the creation of a blackbox system, the state, input and noise dimensions have
+to be defined separately. For a constrained controlled black box system, the macro
```
```suggestion
to be defined separately. For a constrained controlled black-box system, the macro
```

src/macros.jl
```diff
@@ -89,3 +98,501 @@ macro map(ex)
...
+julia> using LazySets
+julia> A = [1. 0; 0 1.]
+julia> B = Matrix([1. 0.5]')
+julia> c = [1., 1.5]
+julia> X = BallInf(zeros(2), 10.)
+julia> U = BallInf(zeros(1), 2.)
+julia> @system(x⁺ = A*x + B*u + c, x∈X, u∈U)
+```
+
+For the creation of a blackbox system, the state, input and noise dimensions have
```
```suggestion
For the creation of a black-box system, the state, input and noise dimensions have
```

src/macros.jl
```diff
@@ -89,3 +98,501 @@ macro map(ex)
...
+            return Meta.parse(str[1:end-1]*"*"*str[end])
+        end
+    end
+end
+
+"""
+    extract_sum(summands, state::Symbol, noise::Symbol)
+
+Given a array of expression `summands` which consists of one or more elements
+which either are mutliplication or a symbol, the corresponding fields of the
```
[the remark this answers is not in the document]
ueliwechsler: Close, it is a multiplication expression or ....

src/macros.jl
ueliwechsler: I agree with you that it is not without peril. I did it this way to have as little verbosity as possible in the macro. 

As an alternative, which I begin to prefer, we could add an input variable, as with state and noise and set it as default to `u` and whenever a different input should be used, it has to be specified, e.g.
`@system( x' = Ax + Bv, input:v)`.

src/macros.jl
```diff
@@ -333,22 +346,50 @@ function extract_dyn_equation_parameters(equation, state, noise, dim, AT)
...
+:(c)
+julia> extract_sum(:(Axi), :xi, :w)
+:(A*x1)
+julia> extract_sum(:(Awb), :xi, :wb)
+:(A*wb)
+julia> extract_sum(:(A1u), :x, :w)
+:(A1*u)
+julia> extract_sum(:(A1ub), :x, :w)
+:(A1u*b)
+```
```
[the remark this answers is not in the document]
ueliwechsler: Thanks. I fixed it. I was sloppy.

ueliwechsler:
Similar to the `noise` default value, I added a default value for `input` `u` which can be changed by adding `input: u1` to the system definition.

src/macros.jl
mforets: Perhaps the wording can be improved; but it means that any symbol can be used. Perhaps with the special treatment of `w`. See the tests for examples.

ueliwechsler: I added the functionality of handling the input in the same way as the noise term to see how it works.
I think it simplifies the documentation, is more natural and it allows also for arbitrary constant terms. Even though it will be more verbose, I think it's worth it.
What do you think?

schillic: > Even though it will be more verbose

Only if you do not use `u`, right?

> What do you think?

I think this is an improvement :+1:

ueliwechsler: > Only if you do not use `u`, right?

Correct.

schillic:
> another try on making doctests work

[You can let `Documenter` fix the doctests for you](https://juliadocs.github.io/Documenter.jl/stable/man/doctests/#Fixing-Outdated-Doctests-1) :wink:

ueliwechsler:
> > another try on making doctests work
> 
> [You can let `Documenter` fix the doctests for you](https://juliadocs.github.io/Documenter.jl/stable/man/doctests/#Fixing-Outdated-Doctests-1) 😉

😂🙈 the more you know

src/macros.jl
mforets: I agree to focus on the desired test set first and overall behavior, then on the implementation details and finally the documentation. Let extend my comments a bi; I'm not asking that you change the implementation, it's to elaborate on my previous comment.

> But if there is a functionality like state_matrix_field(::AbstractSystem) = :A this could be made more robust and indepentend of the fields.

Indeed. That is of independent interest and should be done in  https://github.com/JuliaReach/MathematicalSystems.jl/issues/126. It has not been assigned to anyone yet.

> I tried to make it work like this in the beginning, but then I fought it to be too tedious to implement and not automatically working, therefore, I change to the current approach.

Yes, i think the easiest solution is that each each system defines its template so we can iterate over the templates. Once that is done, the code simplifies a lot:

```julia
julia> @capture(:(x' = A*x + B*u), x_' = A_*x_ + B_*u_)
true

julia> x, A, B, u
(:x, :A, :B, :u)

julia> @capture(:(z' = Q*z + T*w), x_' = A_*x_ + B_*u_)
true

julia> x, A, B, u
(:z, :Q, :T, :w)
```

Regarding discrete systems, the good thing is that they are distinguishable from continuous ones that we can first try `@capture(expr, x_' = rhs_)` (no string parsing is needed); if this fails you have a discrete system (or an error), and the state variable can be detected by finding the occurrence of `⁺` as it is done already in the code.

ueliwechsler:
I tried to rebase from master to add the latest changes, but it doesn't look like it worked. Is there a way to undo this?

src/macros.jl
ueliwechsler: > Yes, i think the easiest solution is that each system defines its template so we can iterate over the templates. Once that is done, the code simplifies a lot:

Did you already figure out an implementation or are you talking about high-level ideas?

In order for the template to work for both continuous and discrete, the template has to be defined without `⁺`and `'` and the dynamic equation needs to be stripped from it first.

The template can then  be used like this
```julia
templ = :(x_ = A_*x_ + B_*u_);
julia> @capture(:(x = A*x + B*u), $templ)
```
where we iterate over all possible types to

schillic:
> Is there a way to undo this?

In principle you can modify every commit, but it requires work. I tried to rebase and got many merge conflicts.

ueliwechsler:
The error handling has to be changed for `@test_throws` to work.
See https://discourse.julialang.org/t/test-throws-with-macros-after-pr-23533/5878/2 for more information.

ueliwechsler:
Using https://github.com/iamed2/ResultTypes.jl could be a good idea.

src/macros.jl
mforets: Hmm i see that `@capture(expr, $temp)` is not good enough as one also needs the matched state symbol and matrices, etc., but `@capture` just returns a boolean.. so one needs a second pass on the expression, or use another macro.. (it would be easier if `@capture` returned a dictionary here, as with `Espresso`).

mforets:
@ueliwechsler thanks for fixing the tests, we are getting closer :+1:

Project.toml
```diff
@@ -4,12 +4,15 @@ version = "0.8.1"
 
 [deps]
 Espresso = "6912e4f1-e036-58b0-9138-08d1e6358ea9"
+InteractiveUtils = "b77e0a4c-d291-57a0-90e8-8db25a27a240"
```
blegat: Can we remove it ?

test/@system.jl
```diff
@@ -0,0 +1,191 @@
...
+    @test @system(x' = -x) == LinearContinuousSystem(-1.0 * Diagonal(ones(1)))
+    @test @system(x' = x, dim=3) == LinearContinuousSystem(Diagonal(ones(3)))
+    @test @system(x' = 2x, dim=3) == LinearContinuousSystem(2.0*Diagonal(ones(3)))
+end
+
+@testset "@system for linear control continuous systems" begin
+    # if the state should be named `w`
+    @test @system(w' = Aw + Bu) == LinearControlContinuousSystem(A, B)
+    # but if the input should be named `w`
+    @test_throws ArgumentError  @system(x' = Ax + Bw)
```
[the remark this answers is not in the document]
mforets: The following test is missing:
```julia
@system(z' = A*z + B*u1)
```
where the right-hand side doesn't contain neither `u` nor `w` terms. At the moment this gives an ERROR (it should at least throw an ArgumentError).

As commented previously, i think that this is case is naturally and should be read as a linear system with input `B*u1`.

Project.toml
ueliwechsler: The `subtypes` method is used here:
https://github.com/JuliaReach/MathematicalSystems.jl/blob/9e523fa0610cbdfeedc9838a5a13a608b8cfa891/src/macros.jl#L106-L110

test/@system.jl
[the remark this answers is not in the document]
ueliwechsler: > As commented previously, i think that this is case is naturally and should be read as a linear system with input `B*u1`.

I have a split opinion on this one. As you know, I did do it this way in the beginning because it is nicer to look at and to use. 

But by using an explicit `input` variable, ambiguities are avoided which makes the code and the explanation more structured and robust.  Additionally, there can be any variable for the constant term `:c` and the behaviour for equations with `*` and without `*` is the same.

ueliwechsler:
I had some more fun with `@system` today and tried to figure out what is possible. 
I added two new functionalities, i.e. handling of special cases:
```julia
# if * is used, an arbitrary variable is consider a input (in this two cases)
@system(w' = A*w + B*u_1) == LinearControlContinuousSystem(A, B)
@system(w' = A*w + B*u_1 + c, w∈X, u_1∈U) == ConstrainedAffineControlContinuousSystem(A, B, c, X, U)

# if there is no input or noise matrix, it is replaced with a Diagonalmatrix
sys = @system(x' = Ax + Bu + w, x∈X, u∈U, w∈W)
sys == NoisyConstrainedLinearControlContinuousSystem(A,B, Diagonal(ones(n)), X, U, W)
sys = @system(x' = Ax + u + w, x∈X, u∈U, w∈W)
sys == NoisyConstrainedLinearControlContinuousSystem(A,Diagonal(ones(n)), Diagonal(ones(n)), X, U, W)
@system(x' = Ax + u) == LinearControlContinuousSystem(A, Diagonal(ones(n)))
@system(x' = Ax + w, x∈X, w∈W) == NoisyConstrainedLinearContinuousSystem(A, Diagonal(ones(n)), X, W)
```
and I found a bug, if the set are defined in arbitrary order, they will be assigned wrongly
```julia
@system(x' = Ax + Bu, u∈U, x∈X) == ConstrainedLinearControlContinuousSystem(A,B,X,U) #false
@system(x' = Ax + Bu, u∈U, x∈X) == ConstrainedLinearControlContinuousSystem(A,B,U,X) #true
```
this could be handled by ordering the set after calling `set = expand_set.(constr, state, input, noise)` in the system macro.

@mforets would it work for you if you have to write `@system(w' = A*w + B*u_1) ` instead of `@system(w' = Aw + Bu_1) `?

mforets:
@ueliwechsler thanks for the update. My comments for each proposed change:

```julia
# if * is used, an arbitrary variable is consider a input (in this two cases)
@system(w' = A*w + B*u_1) == LinearControlContinuousSystem(A, B)
@system(w' = A*w + B*u_1 + c, w∈X, u_1∈U) == ConstrainedAffineControlContinuousSystem(A, B, c, X, U)
```
This is good. In models it is not uncommon to call the input `uᵢₙ`, or whatever other letter. Moreover,
about

> would it work for you if you have to write @system(w' = A*w + B*u_1) instead of @system(w' = Aw + Bu_1) ?

Yup, I agree. lt's not immediate that `Bu_1` should be parsed as `B*u_1` without the extra information `input:u_1`. There can be workarounds, sure, but that's definitely out of the scope of a first version of `@system`. In addition, `@system(w' = Aw + Bu_1, input:u_1)` already works which is quite nice.

```julia
# if there is no input or noise matrix, it is replaced with a Diagonalmatrix
...
@system(x' = Ax + u) == LinearControlContinuousSystem(A, Diagonal(ones(n)))
```
After thinking about it, i'm in favor of this change. On the other hand, can we make this to work? (can be added in a follow-up PR..)

```julia
julia> @system(x' = Ax + u, u ∈ U)
ERROR: ArgumentError: the entry (:A, :B, :U) does not match a `MathematicalSystems.jl` structure
Stacktrace:
 [1] top-level scope at REPL[20]:1
```

> and I found a bug, if the set are defined in arbitrary order, they will be assigned wrongly

It was not clear to me if you fixed it? Also shouldn't the tests read

```julia
@system(x' = Ax + Bu, x∈X, u∈U) == ...
@system(x' = Ax + Bu, u∈U, x∈X) == ...
```

mforets:
A general comment: I always thought we would be using the macro without surrounding parentheses,  as in the comments in [this thread](https://github.com/JuliaReach/MathematicalSystems.jl/issues/76). I find it more comfortable to write `@system x' = Ax + Bu, u ∈ U` instead of `@system(x' = Ax + Bu, u ∈ U)`. After all, the purpose of the macro is to create special syntax that simplifies writing to the 

(Technically, for precedence reasons (?) `x' = Ax + Bu, u ∈ U` is parsed as `x' = (other stuff)` and this has to be considered. But i don't know if there is some other fundamental reason to rule out the version without parentheses. I'd be happy to explore this on a second iteration of `@system`.)

ueliwechsler:
Thanks for the comments.

> Yup, I agree. lt's not immediate that `Bu_1` should be parsed as `B*u_1` without the extra information `input:u_1`. There can be workarounds, sure, but that's definitely out of the scope of a first version of `@system`. In addition, `@system(w' = Aw + Bu_1, input:u_1)` already works which is quite nice.

Good, I am glad you agree. The functionality that allows you to have an arbitrary input is not automatic for all models. I use templates to figure out the input variable from the parsed equation. So far only the templates 
`x_ = A_*x_ + B_*u_`  `x_ = A_*x_ + B_*u_ + c_` are implemented as you can see here
https://github.com/JuliaReach/MathematicalSystems.jl/blob/702732e930e64ced38e814c953f4671c8322cac5/src/macros.jl#L265-L267

I intend to also add `f_(x_,u_)` and not adding any templates where there is a input and noise term. What do you think? Are there other templates that you would add?



> After thinking about it, i'm in favor of this change. On the other hand, can we make this to work? (can be added in a follow-up PR..)
> 
> ```julia
> julia> @system(x' = Ax + u, u ∈ U)
> ERROR: ArgumentError: the entry (:A, :B, :U) does not match a `MathematicalSystems.jl` structure
> Stacktrace:
>  [1] top-level scope at REPL[20]:1
> ```
It should work yes. You need to write `@system(x' = Ax + u, x∈X, u ∈ U)`. But there is still the problem, that since I `esc` the entire Expression. Once you call this, you first have to do `using LinearAlgebra` because it requires that `Diagonal` is in the main namespace. As a quick fix, I will use `MathematicalSystems.Diagonal` such that this is avoided.

> It was not clear to me if you fixed it? Also shouldn't the tests read
> 
> ```julia
> @system(x' = Ax + Bu, x∈X, u∈U) == ...
> @system(x' = Ax + Bu, u∈U, x∈X) == ...
> ```

I did not fix this bug so far. Once I fixed it, I will add the corresponding tests.


> But i don't know if there is some other fundamental reason to rule out the version without parentheses.

As far as I recall, the parsing was just much more annoying without the parantheses which was the reason why I switched.

mforets:
> I intend to also add f_(x_,u_) and not adding any templates where there is a input and noise term. What do you think?

Alright (but in follow-up PR -- let's not add more functionality do this one please).

> You need to write @system(x' = Ax + u, x∈X, u ∈ U).

Good catch, there is no such system without the `x∈X`. As it turns out, in models we often have that the constraint set is not specified (ie. it is the universal set). In the future, i'd like that the universal set is not mandatory (sometimes we have used `LazySets.Universe(n)`, we have also used `nothing`).

> ... you first have to do using LinearAlgebra because it requires that Diagonal is in the main namespace. 

I would suggest that we use here the multiple of the identity matrix, `MathematicalSystems.IdentityMultiple`.

> I did not fix this bug so far. Once I fixed it, I will add the corresponding tests.

If we you postpone this fix to a follow up PR, i won't complain.

> As far as I recall, the parsing was just much more annoying without the parantheses which was the reason why I switched.

Ok. I assume it is harder, but not impossible.

ueliwechsler:
> Ok. I assume it is harder, but not impossible.

Not at all. It should't be to hard, just a bit annoying.

> Alright (but in follow-up PR -- let's not add more functionality do this one please).

Sure. I would change the following in this PR:
- [ ] use `MathematicalSystems.IdentityMultiple` instead of `Diagonal`.

Do you agree with that? Or should I also postphone this?

The following things can be done in a later PR (in approximately increasing order of work needed):
- [ ] add template `f_(x_,u_)` for allowing to write  `@system(x' = f(x,u_12)` (add one line)
- [ ] correct handling for arbitrary order of input set definition + unit_tests
- [ ] add `LazySets.Universe(n)` or  `nothing` as default cases for undefined sets (a straightforward  option would be to handle every set as a `Constrained` set with all set as `Universe(n)` by default and if a set definition is provided specify the corresponding set)
- [ ] Make `@system` macro work without paranthesee, e.g. `@system x' = A*x` (adjust argument parsing)
- [ ] Add support for remaining types, e.g.  `polynomial`
- [ ] Add support for `SystemWithOutput`, e.g. `@system(x' = Ax + Bu, y=Bx + Du)`
- [ ] Allow arbitrary input for dynamic equations without `*`, e.g. `@system(x' = Ax + Bu1)`
- [ ] Major revision by using typ-specific "templates" for each `AbstractSystem` type 

Concerning this PR, what comments should still be addressed? And what can be done from my side?

mforets:
> use MathematicalSystems.IdentityMultiple instead of Diagonal.
> Do you agree with that? 

Ok!

mforets:
> The following things can be done in a later PR (in approximately increasing order of work needed):

Thanks for the write-up, this is very useful.

> Concerning this PR, what comments should still be addressed? And what can be done from my side?

If you agree let's remove the "WIP" in the title, and let's also give the heads up to @schillic and 
@blegat.

src/macros.jl
```diff
@@ -89,3 +99,582 @@ macro map(ex)
...
+    else # summand is parsed as a constant term
+        return summand
+    end
+
+end
+
+"""
+    extract_sum(summands, state::Symbol, input::Symbol, noise::Symbol)
+
+Extract the variable name and field name for the every element of `summands`.
```
```suggestion
Extract the variable name and field name for every element of `summands`.
```

src/macros.jl
```diff
@@ -89,3 +99,582 @@ macro map(ex)
...
+
+If a element of `summands` is a symbol, the symbol is the variable name and the
```
```suggestion
If an element of `summands` is a symbol, the symbol is the variable name and the
```

src/macros.jl
```diff
@@ -89,3 +99,582 @@ macro map(ex)
...
+field name is `:c`. If a element of `summands` is a multiplication expression
```
```suggestion
field name is `:c`. If an element of `summands` is a multiplication expression
```

src/macros.jl
```diff
@@ -89,3 +99,582 @@ macro map(ex)
...
+`lhs*rhs`, return `lhs` as variable name and `:A` as field name if `lhs==state`,
+`:B` as field name if `lhs==input` and `:D` as field name if `lhs==noise`
```
```suggestion
`:B` as field name if `lhs==input` and `:D` as field name if `lhs==noise`.
```

src/macros.jl
```diff
@@ -89,3 +99,582 @@ macro map(ex)
...
+
+
+Given an array of expressions `summands` which consists of one or more elements
+which either are mutliplication expression or symbols.
+
+, the corresponding fields of the
+affine system and the variable are extracted.
```
schillic: This sentence is broken.
Also, neither the docstring nor the function name talk about affine systems so far, so maybe there needs to be more explanation.

src/macros.jl
schillic: What is the status here? The comment still says *everything else*.

src/macros.jl
```diff
@@ -89,3 +99,582 @@ macro map(ex)
...
+                push!(params, (:(IdentityMultiple(1.0*MathematicalSystems.I,$state_dim)), :D))
+            else
+                push!(params, (array, :c))
+            end
+        end
+    end
+    return params
+end
+
+function extract_function(rhs, dim::AbstractVector)
```
schillic: Can you add a small comment to explain what these functions do?

src/macros.jl
```diff
@@ -89,3 +99,582 @@ macro map(ex)
...
+        elseif x == input
+            return  Set, :U
+        elseif x == noise
+            return  Set, :W
+        else
+            error("$expr is not a valid set definition, it does not contain"*
+                  "the state $state, the input $input or noise term $noise")
+        end
+    end
+    throw(ArgumentError("The set-entry $(expr) does not have the correct form `x_ ∈ X_`"))
```
```suggestion
    throw(ArgumentError("the set entry $(expr) does not have the correct form `x_ ∈ X_`"))
```

src/macros.jl
```diff
@@ -89,3 +99,582 @@ macro map(ex)
...
+
+### Note
+
+The `expr` consist of one or several of the following elements:
+- continuous dynamic equation: `x' = Ax`
+- discrete dynamic equation: `x⁺ = Ax`
+- sets: `x ∈ X`
+- dimensionality: `dim: (2,1)` or `dim = 1`
+- defining the input variable: `input: u`, `input = u`
+- defining the noise variable: `noise: w`, `noise = w`
```
schillic: There is no description about `E*x' = ...`.

src/macros.jl
```diff
@@ -89,3 +99,582 @@ macro map(ex)
...
+
+The dynamic equation is parsed as following. The variable on the left hand side
```
```suggestion
The dynamic equation is parsed as follows. The variable on the left-hand side
```

src/macros.jl
```diff
@@ -89,3 +99,582 @@ macro map(ex)
...
+corresponds to the state variable. The input variable is by default `u` and the
+noise variable is by default `w`, if notspecified differently.
```
```suggestion
noise variable is by default `w`, if not specified differently.
```

src/macros.jl
```diff
@@ -89,3 +99,582 @@ macro map(ex)
...
+
+If we want to change the default name of the input or noise variable, this can be
+done by addingthe term `input: var` where `var` corresponds to the new name of
```
```suggestion
done by adding the term `input: var` where `var` corresponds to the new name of
```

src/macros.jl
```diff
@@ -89,3 +99,582 @@ macro map(ex)
...
+
+Let us first create a continuous linear system using this macro:
+
+```jldoctest
+julia> A = [1. 0; 0 1.];
+
+julia> @system(x' = A*x)
+LinearContinuousSystem{Float64,Array{Float64,2}}([1.0 0.0; 0.0 1.0])
+```
+A discrete system can be defined by using  `⁺`:
```
```suggestion
A discrete system can be defined by using `⁺`:
```

test/@system.jl
```diff
@@ -0,0 +1,222 @@
...
+
+    @test @system(x' = A1x + B1u) == LinearControlContinuousSystem(A1, B1)
+    @test @system(z_1' = A*z_1 + B*u_1, input:u_1) == LinearControlContinuousSystem(A, B)
+
+    @test @system(x' = Ax + Bu, x ∈ X, u ∈ U) == ConstrainedLinearControlContinuousSystem(A, B, X, U)
+
+    # if variable in front of input is ommitted add identity matrix
+    @system(x' = Ax + u) == LinearControlContinuousSystem(A, IdentityMultiple(1.0*I,n))
+
+    # if * are used x_ =A_*x_ + B_*u_, u_ is interpreted as input variable,
```
```suggestion
    # if * are used x_ = A_*x_ + B_*u_, u_ is interpreted as input variable,
```

test/@system.jl
```diff
@@ -0,0 +1,222 @@
...
+    # independent of the name used for u_
+    @test @system(w' = A*w + B*u_1) == LinearControlContinuousSystem(A, B)
+    # similarily for x_ =A_*x_ + B_*u_ + c_
```
```suggestion
    # similarily for x_ = A_*x_ + B_*u_ + c_
```

test/@system.jl
```diff
@@ -0,0 +1,222 @@
...
+    sys = @system(x' = f1(x, u, w), x ∈ X, u ∈ U, w ∈ W, dims=(1, 2,3))
+    @test sys == NoisyConstrainedBlackBoxControlContinuousSystem(f1, 1, 2, 3, X, U, W)
+
+    sys = @system(x' = Ax + Bu + Dw, x ∈ X, u ∈ U1, w ∈ W1)
+    @test sys == NoisyConstrainedLinearControlContinuousSystem(A, B, D, X, U1, W1)
+    sys = @system(z_1' = Az_1 + B*v_1 + c + Dd_1, z_1 ∈ X, v_1 ∈ U1, d_1 ∈ W1, input:v_1, d_1:noise)
+    @test sys == NoisyConstrainedAffineControlContinuousSystem(A, B, c, D, X, U1, W1)
+
+
+    # if variable in front of input or noise  is ommitted add identity matrix
```
```suggestion
    # if variable in front of input or noise  is omitted, add identity matrix
```

test/@system.jl
```diff
@@ -0,0 +1,222 @@
...
+
+    # u or other symbols are used for inputs; here u is interpreted as input
+    sys = @system(x⁺ = Ax + Bu, x ∈ X, u ∈ U1)
+    @test sys == ConstrainedLinearControlDiscreteSystem(A, B, X, U1)
+
+    # here v is interpreted as input (if the input has more than one letter, use a *)
+    @system(x⁺ = Ax + B*u123, x ∈ X, u123 ∈ U1, input: u123)
+    @test sys == ConstrainedLinearControlDiscreteSystem(A, B, X, U1)
+
+    # if variable in front of input is ommitted add identity matrix
```
```suggestion
    # if variable in front of input is omitted, add identity matrix
```

src/macros.jl
```diff
@@ -89,3 +99,582 @@ macro map(ex)
...
+    stripped_equation = ""
+    if occursin("'", expr_str)
+        AT = AbstractContinuousSystem
+        stripped_equation =  replace(expr_str, "'" => "")
+    elseif occursin("⁺", expr_str)
+        AT = AbstractDiscreteSystem
+        stripped_equation =  replace(expr_str, "⁺" => "")
+    end
+    stripped_expr = Meta.parse(stripped_equation)
+    !@capture(stripped_expr, lhs_ = rhs_) && return (nothing, nothing, nothing)
```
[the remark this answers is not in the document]
ueliwechsler: No, I cannot directly use `is_equation` because `lhs` and `rhs` are extract from the expression `stripped_expr` and used in the next line.

src/macros.jl
```diff
@@ -89,3 +99,582 @@ macro map(ex)
...
+
+    if @capture(lhs, (E_*x_)) || @capture(lhs, (x_))
```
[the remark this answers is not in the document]
ueliwechsler: I would not consider it ignoring, it allows for the `lhs` to additionally have the structure `E*x` where `E` can be any variable name.
We do not need the explicit value of `E` here, it is only needed to extract the name of the state variable.
I added some more comments in the source code.

src/macros.jl
ueliwechsler: This was a left-over from earlier times. The status is that there are default cases for `state`, `input` and `noise` (`:x`, `:u` and `:w`), if you want to parse a different variable name in general, you need to add it additionally in the form `input:u_1` or `input=u_1` as an example where the input should be `u_1`.

Additionally, if the rhs of the equation provided in the macro has the form  `A_*x_ + B_*u_` or `A_*x_ + B_*u_ + c_` the variable `U` will be parsed as input independent of its entry. If it has the same name as the noise, it will still be parsed as input( because in the if conditions, we first check for equality of the variable with the input before checkign for equality with the noise).

src/macros.jl
ueliwechsler: I tried yes.

src/macros.jl
ueliwechsler: True. I added a paragraph in the docstring.

src/macros.jl
ueliwechsler: I remove it and added a hopeful more comprehensive explanation

src/macros.jl
```diff
@@ -89,3 +99,582 @@ macro map(ex)
...
+        summands = add_asterisk.([A, B...], Ref(state), Ref(input), Ref(noise))
+        push!(rhs_params, extract_sum(summands, state, input, noise)...)
+    elseif @capture(rhs, f_(a__))  && f != :(*) && f != :(-) # If rhs is function call
+        # the dimension argument needs to be a iterable
+        (dim == nothing) && throw(ArgumentError("for a blackbox system, the dimension has to be defined"))
+        dim_vec = [dim...]
+        push!(rhs_params, extract_function(rhs, dim_vec)...)
+    else # if rhs is a single term (eith A*x, Ax, 2x, x or 0)
+        # assume dim = 1 if not specified
+        dim = (dim == nothing) ? 1 : dim
```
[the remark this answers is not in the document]
ueliwechsler: The reasoning behind this (or the be more exact, the happenstances) is that this `dim=1` as default is only defined for system that have a single term on the right-hand side, i.e. `x'=0`, `x=x` , `x=2x` and `x=Ax`. 
In the cases `x=0` and `x=2x` it feels natural to assume that it corresponds to the scalar case and if not you have to write `x=0, dim=2` to specify otherwise. And for sure, any other default cases which is not `dim=1` is  more arbitrary.

Another argument for a default case is that the `dim` definition is only needed for the the black-box system, because in this case, there is no additional information provided in contrast to the affine case where we have normally the matrices available.

However, thinking about his now, I don't feel that strongly towards having `dim=1` as default case. Even though, It is  more work to add the multiple error messages 😉

ueliwechsler:
I tried to address the comments from @schillic as well as possible and updated the todo list https://github.com/JuliaReach/MathematicalSystems.jl/pull/125#issuecomment-575240459.

I found some more "loopholes" in the definition and the documentation is not yet where it should be, but I think the behavior of the macro, in general, is as it is supposed to be. The requested or intended changes are either bug fixes or additional feature, but not breaking.

The remaining design decision that comes to my mind is if we want to have `dim=1` as default case for affine systems with only one term such as `x'=0` or `x'=2x`.

mforets:
> The remaining design decision that comes to my mind is if we want to have dim=1 as default case for affine systems with only one term such as x'=0 or x'=2x.

If the scalar differential equation `x' = 2x` can be written as `@system x' = 2x`, that's *simple* and *intuitive*. If we require to additionally write `, dim=1`, this adds one (small but) unnecessary level of verbosity. The argument `dim` allows one to specify a non-default value of the state-space dimension.

src/macros.jl
```diff
@@ -89,3 +99,610 @@ macro map(ex)
...
+        return summand
+    end
+
+    str = string(summand)
+    statestr = string(state); lenstate = length(statestr)
+    inputstr = string(input); leninput = length(inputstr)
+    noisestr = string(noise); lennoise = length(noisestr)
+
+    # if summand contains the state, input or noise variable at the end and has
+    # ond or more additional characters (i.e. length(str) > length(state) for the state)
```
```suggestion
    # one or more additional characters (i.e. length(str) > length(state) for the state)
```

src/macros.jl
```diff
@@ -89,3 +99,610 @@ macro map(ex)
...
+    # a `*` is added inbetween.
```
```suggestion
    # a `*` is added in between
```

src/macros.jl
```diff
@@ -89,3 +99,610 @@ macro map(ex)
...
+corresponds to the elements of the rhs of an affine system.
+
+If an element of `summands` is a multiplication expression
+`lhs*rhs`, return `lhs` as variable name and `:A` as field name if `lhs==state`,
+`:B` as field name if `lhs==input` and `:D` as field name if `lhs==noise`.
+
+If an element of `summands` is a symbol, and not equal to `input` or `noise`,
+the symbol is the variable name and the field name is `:c`. If it is equal to
+`input`, the variable name is a `IdentityMultiple(I,state_dim)` where `state_dim`
+is extracted from the variable name of the state and the field name is `:B`.
```
schillic: I do not understand how you can extract the dimension from a name.

src/macros.jl
```diff
@@ -89,3 +99,610 @@ macro map(ex)
...
+                push!(params, (:(IdentityMultiple(1.0*MathematicalSystems.I,$state_dim)), :D))
+            else
+                push!(params, (array, :c))
+            end
+        end
+    end
+    return params
+end
+
+# Extract the variable name of the function of the rhs of a equation if it has
```
```suggestion
# Extract the variable name of the function of the rhs of an equation if it has
```

src/macros.jl
```diff
@@ -89,3 +99,610 @@ macro map(ex)
...
+# the form `f_(x_)`, `f_(x_,u_)` or `f_(x_,u_,w_)`.
+# In addition to the variable name of the function and the field name `:f`,
+# depening on the number of input arguments, the field names `:statedim`, `:inputdim`
```
```suggestion
# depending on the number of input arguments, the field names `:statedim`, `:inputdim`
```

src/macros.jl
```diff
@@ -89,3 +99,610 @@ macro map(ex)
...
+                         (dim[2], :inputdim)]
+    elseif @capture(rhs, f_(x_,u_,w_))
+        @assert length(dim) == 3
+        return [(f, :f), (dim[1], :statedim),
+                         (dim[2], :inputdim),
+                         (dim[3], :noisedim)]
+    end
+end
+
+# Take the vectors of tupples providing the variable and fields names for the
```
```suggestion
# Take the vectors of tuples providing the variable and fields names for the
```

src/macros.jl
```diff
@@ -89,3 +99,610 @@ macro map(ex)
...
+# lhs, the rhs and the sets combined them and return a vector of field names
+ # and a vector of variable names.
```
```suggestion
# and a vector of variable names.
```

src/macros.jl
```diff
@@ -89,3 +99,610 @@ macro map(ex)
...
+    elseif @capture(rhs, f_(x_,u_,w_))
+        @assert length(dim) == 3
+        return [(f, :f), (dim[1], :statedim),
+                         (dim[2], :inputdim),
+                         (dim[3], :noisedim)]
+    end
+end
+
+# Take the vectors of tupples providing the variable and fields names for the
+# lhs, the rhs and the sets combined them and return a vector of field names
```
schillic: Not sure if this is what you mean:
```suggestion
# lhs, the rhs and the sets, combine them, and return a vector of field names
```

src/macros.jl
```diff
@@ -89,3 +99,610 @@ macro map(ex)
...
+    rhs_var_names = [tuple[1] for tuple in rhs]
+    lhs_var_names = [tuple[1] for tuple in lhs]
+    set_var_names = [tuple[1] for tuple in set]
+    field_names = (rhs_fields..., lhs_fields..., set_fields...)
+    var_names = (rhs_var_names..., lhs_var_names..., set_var_names...)
+    return field_names, var_names
+end
+
+# extract the variable name and the field name for a set expression. The method
+# check wether the set belongs to the `state`, `input` or `noise` and returns a
```
```suggestion
# checks wether the set belongs to the `state`, `input` or `noise` and returns a
```

src/macros.jl
```diff
@@ -89,3 +99,610 @@ macro map(ex)
...
+# tupple of symbols where the field name is either `:X`, `:U` or `:W`  and
```
```suggestion
# tuple of symbols where the field name is either `:X`, `:U` or `:W` and
```

src/macros.jl
```diff
@@ -89,3 +99,610 @@ macro map(ex)
...
+# the variable name is the value parse as Set_.
```
```suggestion
# the variable name is the value parsed as Set_
```

src/macros.jl
```diff
@@ -89,3 +99,610 @@ macro map(ex)
...
+
+The dynamic equation is parsed as follows. The variable on the left-hand side
+corresponds to the state variable. The input variable is by default `u` and the
+noise variable is by default `w`, if not specified differently.
+
+If we want to change the default name of the input or noise variable, this can be
+done by adding the term `input: var` where `var` corresponds to the new name of
+the input variable.
+
+As a exception to the rule, if the right-hand side has the form `A*x + B*u` or
```
```suggestion
As an exception to the rule, if the right-hand side has the form `A*x + B*u` or
```

src/macros.jl
ueliwechsler: It is done by taking the symbol corresponding to the already defined state matrix which is stored in the variable `array` and generating the expression `state_dim =  Expr(:call, :size, :($array), 1)` which is evaluated in the scope where `@system` is called. I will try to explain it in more detail in the comment.

src/macros.jl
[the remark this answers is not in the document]
ueliwechsler: I am not sure about that. It would be good for consistency of handling `dim` for sure.
We would have `dim=(1,1,1)` as the default value for the state, input and noise. But if the function that the user provides does not have the same dimensions, the macro would just generate a "wrong" system since the dimensions do not match the function arguments. 
(this is not the case for e.g. `x=2x`, there the macro may create a different system than the user expected but it is "correct")

ueliwechsler:
> If the scalar differential equation `x' = 2x` can be written as `@system x' = 2x`, that's _simple_ and _intuitive_

I agree with you. The question came up in a review comment.

> The argument dim allows one to specify a non-default value of the state-space dimension.

Another point as part of a comment is having similar default value for `black-box` sytems, i.e. `dim=(1,1,1)` as default. What is your opinion on that?

src/macros.jl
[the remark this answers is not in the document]
schillic: Okay, thanks for the explanations. Then we can keep it as is.

mforets:
> Another point as part of a comment is having similar default value for black-box sytems, i.e. dim=(1,1,1) as default (see #125 (comment)). What is your opinion on that?

In such systems, the dimension is part of the struct so providing that information seems reasonable. Moreover, how can one tell the dimension from `x ' = f(x)`? I can't think of an example where there could be an ambiguity. In `x' = 2x`, if `x` is a vector we can pass `dim = n` otherwise it is interpreted as a scalar.

ueliwechsler:
I agree. So we keep the implementation as it is now, right?
Is there anything else to address for this PR?

schillic:
> Is there anything else to address for this PR?

There are two unresolved discussions [here](https://github.com/JuliaReach/MathematicalSystems.jl/pull/125#discussion_r361145923) and [here](https://github.com/JuliaReach/MathematicalSystems.jl/pull/125#discussion_r368890274).

src/macros.jl
ueliwechsler: @mforets what's your take on the templates? Do you agree to implement them in a separate PR?

mforets:
LGTM, great job!!! :tada:

Regarding the other tasks listed before, I opened https://github.com/JuliaReach/MathematicalSystems.jl/issues/133.

Issue #128, opened 2019-12-28
Rename affine term in map fields


mforets:
Follow up of https://github.com/JuliaReach/MathematicalSystems.jl/pull/124 for `map`.

schillic:
What is the plan here? The affine term is already called `c`.

mforets:
For example in `AffineMap` its called b

PR #129, opened 2019-12-28
Minor version bump

mforets:


schillic:
I suggest to do the breaking change in #128 in the same version.

mforets:
> I suggest to do the breaking change in #128 in the same version.

True, it would put 2 related changes on the same release, it's more organized than having them separated in 2 releases, I agree. On the other hand, that involves work that hasn't been done yet. In our development model, I'm on the camp of having small but frequent release cycles.

schillic:
Okay, I'll do the change shortly.

mforets:
Thanks!

PR #145, opened 2020-01-30
enable variable as kwargs in @map for identity map

### LINKED ISSUE #115: @map macro with kwargs fails for variables
```julia
julia> using MathematicalSystems

julia> dim = 3
julia> @map( x -> x, dim = 3)
julia> @map( x -> x, dim = dim) #fails
ERROR: UndefVarError: #43#dim not defined
Stacktrace:
 [1] top-level scope at C:\Users\wueli\.julia\packages\MathematicalSystems\5vm3X\src\macros.jl:59
```
Same as in #113  but with the `macro(ex, arg)` signature.

ueliwechsler:
closes #115

### DIFF
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -60,8 +60,8 @@ macro map(ex, args)
     dimension = nothing
     x = (ex.args)[1]
     rhs = (ex.args)[2].args[2]
-    @show args
     if x == rhs
+        # identity map: x -> x, dim=..
         if @capture(args, (dim = dim_) | (dim: dim_) )
             dimension = dim
         else
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -57,18 +57,20 @@ IdentityMap(5)
 ```
 """
 macro map(ex, args)
-    quote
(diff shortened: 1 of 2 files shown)


src/macros.jl
```diff
@@ -57,18 +57,20 @@ IdentityMap(5)
...
-        MT = IdentityMap
-        local pat = Meta.parse("$x")
-        local matched = matchex(pat, rhs)
-        matched != nothing && return MT($args)
-
-        throw(ArgumentError("unable to match the given expression to a known map type"))
+    dimension = nothing
+    x = (ex.args)[1]
+    rhs = (ex.args)[2].args[2]
+    @show args
```
```suggestion
```

src/macros.jl
```diff
@@ -57,18 +57,20 @@ IdentityMap(5)
...
+    if x == rhs
+        if @capture(args, (dim = dim_) | (dim: dim_) )
```
```suggestion
        # identity map: x -> x, dim=..
        if @capture(args, (dim = dim_) | (dim: dim_) )
```

PR #147, opened 2020-01-30
Parse IVPs with @system

### LINKED ISSUE #165: Extend @system to handle initial-value problems
This is one concrete idea that has been proposed in https://github.com/JuliaReach/MathematicalSystems.jl/issues/137.

mforets:
Closes #165.

### DIFF
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -745,7 +745,7 @@ macro system(expr...)
         if x0 == nothing
             return esc(sys)
         else
-            return Expr(:call, InitialValueProblem, esc(:($(sys))), esc(:($(x0))))
+            return Expr(:call, InitialValueProblem, esc(:($sys)), esc(:($x0)))
         end
     catch ex
         if  isa(ex, ArgumentError)
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -244,7 +244,6 @@ function strip_dynamic_equation(expr)
 end
 
 function parse_system(exprs)
-
     # define default dynamic equation, unknown abstract system type,
     # and empty list of constraints
     dynamic_equation = nothing
@@ -256,6 +255,7 @@ function parse_system(exprs)
     input_var = :u
     noise_var = :w
     dimension = nothing
+    initial_state = nothing # for initial-value problems
 
     # main loop to parse the subexpressions in exprs
     for ex in exprs
@@ -288,6 +288,14 @@ function parse_system(exprs)
                                     "see the documentation for valid examples"))
             end
 
+        elseif @capture(ex, x_(0) ∈ X0_)
+            # TODO? handle equality, || @capture(ex, x_(0) = X0_)
+            if x != state_var
+                throw(ArgumentError("the initial state assignment, $x(0), does "*
+                "not correspond to the state variable $state_var"))
+            end
+            initial_state = X0
+
         elseif @capture(ex, state_ ∈ Set_)  # parse a constraint
             push!(constraints, ex)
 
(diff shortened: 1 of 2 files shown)


src/macros.jl
```diff
@@ -288,6 +288,14 @@ function parse_system(exprs)
                                     "see the documentation for valid examples"))
             end
 
+        elseif @capture(ex, x_(0) ∈ X0_)
+            # TODO? handle equality, || @capture(ex, x_(0) = X0_)
```
schillic: I do not understand what this comment means.

src/macros.jl
```diff
@@ -721,16 +732,21 @@ ConstrainedBlackBoxControlDiscreteSystem{typeof(f),BallInf{Float64},BallInf{Floa
...
         set = extract_set_parameter.(constr, state, input, noise)
         # TODO: order set variables such that the order is X, U, W
         field_names, var_names = constructor_input(lhs, rhs, set)
         sys_type = _corresponding_type(AT, field_names)
-        return  esc(Expr(:call, :($sys_type), :($(var_names...))))
+        sys = Expr(:call, :($sys_type), :($(var_names...)))
+        if x0 == nothing
+            return esc(sys)
+        else
+            return Expr(:call, InitialValueProblem, esc(:($(sys))), esc(:($(x0))))
```
```suggestion
            return Expr(:call, InitialValueProblem, esc(:($sys)), esc(:($x0)))
```

schillic: Previously the `return` statement was `return esc(Expr(...))` and now it is only `return Expr(...)`. That looks incompatible. Can you not just call the `InitialValueProblem` constructor in the normal way?

src/macros.jl
mforets: The "singleton" case, `@ivp(x' = -x, x(0) = x0)`. I think that we can handle it in the clause that checks for an equation, but i didn't check too much; i would rather leave this for another PR.

src/macros.jl
mforets: Maybe, i don't know, what do you mean by the normal way? Returning an expression call is how the map macro was fixed for some use case, see https://github.com/JuliaReach/MathematicalSystems.jl/pull/114/files#diff-0033f9de61e25a98b730785191eb2c68R88

ueliwechsler: I think the comment is about returning 
```julia
Expr(:call, InitialValueProblem, esc(:($sys)), esc(:($x0)))
```
instead of 
```julia
esc(Expr(:call, InitialValueProblem, :($sys), :($x0))
```
I think, it does not matter if we `esc` the entire expression or only the argument, since `InitialValueProblem` is in both "namespaces" available.

schillic: > what do you mean by the normal way?

Just `InitialValueProblem(...)`. But that probably does not work.

src/macros.jl
```diff
@@ -721,16 +732,21 @@ ConstrainedBlackBoxControlDiscreteSystem{typeof(f),BallInf{Float64},BallInf{Floa
...
         # TODO: order set variables such that the order is X, U, W
         field_names, var_names = constructor_input(lhs, rhs, set)
         sys_type = _corresponding_type(AT, field_names)
-        return  esc(Expr(:call, :($sys_type), :($(var_names...))))
+        sys = Expr(:call, :($sys_type), :($(var_names...)))
+        if x0 == nothing
+            return esc(sys)
+        else
+            return Expr(:call, InitialValueProblem, esc(:($sys)), esc(:($x0)))
+        end
```
schillic: Again my question from [here](https://github.com/JuliaReach/MathematicalSystems.jl/pull/147#discussion_r378121412), just with everything in the view.

src/macros.jl
ueliwechsler: I think we tried it and as a result you could not `esc` the variables.

src/macros.jl
ueliwechsler: I checked. Both ways should be fine.
For consistency, we could use
```julia
ivp = Expr(:call, InitialValueProblem, :($sys), :($x0))
return esc(ivp)
```

mforets:
Thanks for your reviews.

Issue #137, opened 2020-01-27
Add @ivp macro


mforets:
Proposal: add `@ivp` for initial-value problem, that behaves like `@system` but additionally receives the initial condition, eg.
 
```julia
julia> @ivp(x' = -x, x(0) ∈ Interval(-1.0, 1.0))
```

ueliwechsler:
What about extending the `@system` macro with an additional property? Which does create a `IVP` if there is a `state(0)  ∈  set` structure?
```julia
julia> @system(x' = -x, x(0) ∈ Interval(-1.0, 1.0))
```

mforets:
Ok, good idea :+1: .  `InitialValueProblem` are also systems.

We can extend `@system` as a first step, but i still consider adding `@ivp`, macro which *requires* `x(0) ∈  ...` to be present. Usage within reachability problems would be:

```julia
P = @ivp x' = -x, x(0) ∈ Interval(-1.0, 1.0)
solve(P, T=1.0)
```

ueliwechsler:
So the `@ivp` macro would basically be `@system` with an additional check, for the property `state(0)`, right?

mforets:
Yes, that's exactly what i have in mind so far. So we can re-use as much as the existing functionality of `@system` as possible.

schillic:
If you want to parse an expression as `S, I`, it mainly depends on how the parser binds the comma.
`@icp x' = Ax, X, Y` can be parsed as either
* `S` being `x' = Ax, X` and `I` being `Y` (that is what you want here) or
* `S` being `x' = Ax` and `I` being `X, Y`.

mforets:
Not sure i'm following -- what is `I` here? For what it's worth, "native" Julia parsing for `@ivp x' = Ax, X, Y` is an assignment operation, `x' = (....)`.

schillic:
> what is I here?

My impression is that you want `@ivp` to match the pattern `S, I` where `S` is a system definition and `I` is a definition of the initial states.

mforets:
Ok. I thought about that too, it could be useful. But for `@ivp` i was proposing to start with

```julia
P = @ivp x' = -x, x(0) ∈ Interval(-1.0, 1.0)
```
which would re-use the `@system` stuff.

We can try to make this work in a later stage:

```julia
S = @system x' = x 
P = @ivp S, x(0) ∈ Interval(-1.0, 1.0)
```

schillic:
[...]

PR #148, opened 2020-01-30
Create @ivp macro

mforets:
Closes https://github.com/JuliaReach/MathematicalSystems.jl/issues/137


mforets:
The idea is just to re-use the code for `@system` but return an error if `x(0) ... ` was not defined.

ueliwechsler:
Can I push a proposal directly into this branch?

mforets:
Yes, go ahead. When i get to the office i can review it. BTW yesterday i played around with `@ivp`, see the branch mforets/ivp_macro in this repo.

ueliwechsler:
I like your version, would have done something similar. So no need to write my own proposal.
I just added a negative test for `@ivp` without specified `x(0)`.

One comment on your `mforets/ivp_macro`. I think you can throw the error like this 
` throw(ArgumentError("..."))` instead of  `:(throw("..."))`

mforets:
> I like your version, would have done something similar. So no need to write my own proposal.

Ok, thanks for taking a look! I'll push it here then.

> I just added a negative test for @ivp without specified x(0).

Good idea, thanks.

> I think you can throw the error like this
throw(ArgumentError("...")) instead of :(throw("..."))

Alright, thanks for the tip.

mforets:
This PR is read for review. Tests pass locally.

PR #150, opened 2020-01-31
#90 - Overload more arithmetic operations for IdentityMultiple 

### LINKED ISSUE #90: Add missing arithmetic operations of IdentityMultiple

### LINKED ISSUE #149: divide IdentityMultiple by UniformScaling throws error
```julia
IdentityMultiple(I, 2) / I
ERROR: MethodError: no method matching -(::CartesianIndex{2}, ::Int64)
Closest candidates are:
  -(::Complex{Bool}, ::Real) at complex.jl:303
  -(::Missing, ::Number) at missing.jl:115
  -(::Base.CoreLogging.LogLevel, ::Integer) at logging.jl:107
  ...
Stacktrace:
 [1] getindex(::IdentityMultiple{Bool}, ::CartesianIndex{2}) at C:\Users\wueli\.julia\dev\MathematicalSystems\src\identity.jl:83
 [2] _broadcast_getindex at .\broadcast.jl:596 [inlined]
 [3] _getindex at .\broadcast.jl:626 [inlined]
 [4] _broadcast_getindex at .\broadcast.jl:602 [inlined]
 [5] getindex at .\broadcast.jl:563 [inlined]
 [6] macro expansion at .\broadcast.jl:909 [inlined]
 [7] macro expansion at .\simdloop.jl:77 [inlined]
 [8] copyto! at .\broadcast.jl:908 [inlined]
 [9] copyto! at .\broadcast.jl:863 [inlined]
 [10] copy at .\broadcast.jl:839 [inlined]
 [11] materialize at .\broadcast.jl:819 [inlined]
 [12] broadcast_preserving_zero_d at .\broadcast.jl:808 [inlined]
 [13] /(::IdentityMultiple{Bool}, ::Bool) at .\arraymath.jl:55
 [14] /(::IdentityMultiple{Bool}, ::UniformScaling{Bool}) at .\none:1
 [15] top-level scope at none:0
```
This should be fixable by overlading `:(/)` with something like:
```julia
julia> Base.:(/)(x::IdentityMultiple, i::LinearAlgebra.UniformScaling) = x/i.λ
```

mforets:
Closes #90.
Closes #149.

### DIFF
--- a/src/identity.jl
+++ b/src/identity.jl
@@ -52,7 +52,7 @@ IdentityMultiple{Rational{Int64}} of value 4//1 and order 2
 ```
 
 To create the matrix with a value different from the default (`1.0`), there are
-two ways. Either pass the value through the callable `I`, as in
+two ways. Either pass the value through the callable `I`, as in:
 
 ```jldoctest identitymultiple
 julia> I2 = I(2.0, 2)
@@ -62,7 +62,7 @@ julia> I2r = I(2//1, 2)
 IdentityMultiple{Rational{Int64}} of value 2//1 and order 2
 ```
 
-Or use the lower level constructor passing the `UniformScaling` (`I`):
+Or use the constructor passing the `UniformScaling` (`I`):
 
 ```jldoctest identitymultiple
 julia> I2 = IdentityMultiple(2.0*I, 2)
@@ -80,9 +80,9 @@ end
 Base.IndexStyle(::Type{<:IdentityMultiple}) = IndexLinear()
 Base.size(𝐼::IdentityMultiple) = (𝐼.n, 𝐼.n)
 Base.getindex(𝐼::IdentityMultiple, inds...) = getindex(𝐼.M, inds...)
-Base.getindex(𝐼::IdentityMultiple{T}, ind) where {T} =
-    rem(ind-1, 𝐼.n+1) == 0 ? 𝐼.M.λ : zero(T)
-Base.setindex!(𝐼::IdentityMultiple, X, inds...) = error("cannot store a value in an `Identity`")
+Base.getindex(𝐼::IdentityMultiple{T}, ind) where {T} = rem(ind-1, 𝐼.n+1) == 0 ? 𝐼.M.λ : zero(T)
+Base.setindex!(𝐼::IdentityMultiple, X, inds...) = error("cannot store a value in "*
+                "an `IdentityMultiple`, because this type is immutable")
 
 Base.:(*)(x::Number, 𝐼::IdentityMultiple) = IdentityMultiple(x * 𝐼.M, 𝐼.n)
 Base.:(*)(𝐼::IdentityMultiple, x::Number) = IdentityMultiple(x * 𝐼.M, 𝐼.n)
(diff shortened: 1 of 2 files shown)


src/identity.jl
```diff
@@ -101,21 +103,52 @@ function Base.:(*)(A::AbstractMatrix, 𝐼::IdentityMultiple)
...
+    return IdentityMultiple(U * inv(𝐼.M.λ), 𝐼.n)
+end
+
 function Base.show(io::IO, ::MIME"text/plain", 𝐼::IdentityMultiple{T}) where T
     print(io, "$(typeof(𝐼)) of value $(𝐼.M.λ) and order $(𝐼.n)")
 end
 
-# callable identity matrix
+# callable identity matrix given the size and the numeric type
 LinearAlgebra.I(n::Int, N=Float64) = IdentityMultiple(one(N)*I, n)
```
ueliwechsler:
```suggestion
 LinearAlgebra.I(n::Int, N::DataType=Float64) = IdentityMultiple(one(N)*I, n)
```
to avoid that `I(3,3)` throws a ambiguity error and correct behavior of `I(3,3.0)`

src/identity.jl
```diff
@@ -80,11 +80,13 @@ end
 Base.IndexStyle(::Type{<:IdentityMultiple}) = IndexLinear()
 Base.size(𝐼::IdentityMultiple) = (𝐼.n, 𝐼.n)
 Base.getindex(𝐼::IdentityMultiple, inds...) = getindex(𝐼.M, inds...)
-Base.getindex(𝐼::IdentityMultiple{T}, ind) where {T} =
-    rem(ind-1, 𝐼.n+1) == 0 ? 𝐼.M.λ : zero(T)
-Base.setindex!(𝐼::IdentityMultiple, X, inds...) = error("cannot store a value in an `Identity`")
+Base.getindex(𝐼::IdentityMultiple{T}, ind) where {T} = rem(ind-1, 𝐼.n+1) == 0 ? 𝐼.M.λ : zero(T)
```
ueliwechsler: A check if `ind<I.n` could make sense, to avoid that  `I(3)[1,4]` returns `0`.

src/identity.jl
mforets: Good catch, thanks.

src/identity.jl
mforets: Good idea -- i added bounds checks and `BoundsError` in case of error. The tuple case, `inds...`, has to be special cased, because relying on `LinearAlgebra.UniformScaling` is not good, as that case doesn't check the size (because it doesn't know the size). So i don't know if this is the most efficient way, but it can be re-written like:

```julia
function Base.getindex(𝐼::IdentityMultiple, inds...)
    any(idx -> idx > 𝐼.n, inds) && throw(BoundsError(𝐼, inds))
    getindex(𝐼.M, inds...)
end
```

src/identity.jl
```diff
@@ -80,11 +80,13 @@ end
...
+Base.setindex!(𝐼::IdentityMultiple, X, inds...) = error("cannot store a value in "*
+                "an `IdentityMultiple`, because this type is immutable")
```
schillic:
```suggestion
                "an `IdentityMultiple` because this type is immutable")
```

src/identity.jl
mforets: Actually for `Base.getindex(𝐼::IdentityMultiple{T}, ind)` it is wrong to check `n` because linear indexing implies that the index in the successive columns wrap into one vector; the check should be up to `n^2`. For example, `rand(2, 2)[4]` makes sense but `rand(2, 2)[5]` does not.

PR #151, opened 2020-01-31
#144 - Add getter function for x0 in ivp

### LINKED ISSUE #144: Getter function for initial state in IVP
Proposal: add a new getter function for the `x0` field (initial state) in an initial-value problem. Possible name: `initial_state`, or `initstate`.

mforets:
Closes #144.

### DIFF
--- a/src/ivp.jl
+++ b/src/ivp.jl
@@ -1,8 +1,8 @@
 """
-    InitialValueProblem
+    InitialValueProblem{S <: AbstractSystem, XT} <: AbstractSystem
 
 Parametric composite type for initial value problems. It is parameterized in the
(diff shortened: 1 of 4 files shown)


src/ivp.jl
```diff
@@ -40,6 +45,21 @@ islinear(ivp::InitialValueProblem) = islinear(ivp.s)
...
+
+Return the initial state of an initial-value problem.
+
+### Input
+
+- `ivp` -- initial-value problem
+
+### Output
+
+The initial state field in an initial-value problem.
```
schillic:
```suggestion
The initial state of an initial-value problem.
```

PR #152, opened 2020-01-31
#126 - Add getter functions for systems' field names

### LINKED ISSUE #126: Add getter functions for field names
_Originally posted by @ueliwechsler in https://github.com/JuliaReach/MathematicalSystems.jl/pull/124#issuecomment-568711999_

### LINKED ISSUE #127: Document naming convention
https://github.com/JuliaReach/MathematicalSystems.jl/pull/124#issuecomment-568709464

Functions to be added:

- `state_matrix`
- `input_matrix`
- `noise_matrix`
- `affine_term`

mforets:
Closes #126.
Closes #127.

### DIFF
--- a/src/abstract.jl
+++ b/src/abstract.jl
@@ -174,7 +174,7 @@ function apply end
 """
     state_matrix(::AbstractSystem)
 
-Return the state matrix of a system.
+Return the state matrix of an affine system.
 
 ### Notes
 
--- a/src/abstract.jl
+++ b/src/abstract.jl
@@ -198,7 +198,7 @@ input_matrix(::AbstractSystem)
 """
     noise_matrix(::AbstractSystem)
 
-Return the noise matrix of a system with noise.
+Return the noise matrix of a system with linear noise.
 
 ### Notes
 
--- a/src/abstract.jl
+++ b/src/abstract.jl
@@ -170,3 +170,50 @@ function isaffine(::AbstractMap) end
 Apply the rule specified by the map to the given arguments.
 """
 function apply end
+
+"""
+    state_matrix(::AbstractSystem)
+
+Return the state matrix of a system.
+
+### Notes
+
+The state matrix is the matrix proportional to the state, e.g. the matrix `A`
+in the linear continuous system ``x' = Ax``.
+"""
+state_matrix(::AbstractSystem)
+
+"""
+    input_matrix(::AbstractSystem)
+
+Return the input matrix of a system with input.
+
+### Notes
+
+The input matrix is the matrix proportional to the input, e.g. the matrix `B`
+in the linear continuous system with input, ``x' = Ax + Bu``.
+"""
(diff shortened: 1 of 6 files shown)


src/abstract.jl
```diff
@@ -170,3 +170,50 @@ function isaffine(::AbstractMap) end
 Apply the rule specified by the map to the given arguments.
 """
 function apply end
+
+"""
+    state_matrix(::AbstractSystem)
+
+Return the state matrix of a system.
```
ueliwechsler: Maybe we could write affine system.

mforets: > Maybe we could write affine systems in the docstring

Huh i didn't understand, can you elaborate? Do you propose to extend the docstring of the generic `state_matrix` to sth like:

```julia
In the case of affine systems of the form, ``x' = Ax + c``, the state matrix is ``A``.
```

ueliwechsler: I meant that we write in the docstrinr
"Return the state matrix of an affine system"

schillic: I kind of agree that this function only makes sense for affine systems (as opposed to polynomial systems etc.).

src/abstract.jl
```diff
@@ -170,3 +170,50 @@ function isaffine(::AbstractMap) end
...
+
+The state matrix is the matrix proportional to the state, e.g. the matrix `A`
+in the linear continuous system ``x' = Ax``.
+"""
+state_matrix(::AbstractSystem)
+
+"""
+    input_matrix(::AbstractSystem)
+
+Return the input matrix of a system with input.
```
schillic:
```suggestion
Return the input matrix of a system with linear input.
```

src/abstract.jl
```diff
@@ -170,3 +170,50 @@ function isaffine(::AbstractMap) end
...
+
+The input matrix is the matrix proportional to the input, e.g. the matrix `B`
+in the linear continuous system with input, ``x' = Ax + Bu``.
+"""
+input_matrix(::AbstractSystem)
+
+"""
+    noise_matrix(::AbstractSystem)
+
+Return the noise matrix of a system with noise.
```
schillic:
```suggestion
Return the noise matrix of a system with linear noise.
```

schillic:
> Closes #127.

There is no documentation of the naming convention here. This is important for some of the meta programming to work, so I would not close that issue yet.

mforets:
> There is no documentation of the naming convention here.

True; i added a paragraph about the naming convention in the last commit.

> This is important for some of the meta programming to work, so I would not close that issue yet.

Yes, although i would prefer that internally we also use the getter functions (this is somewhere in the TODO list for `@system`). In that way, any extension that provides the required getter functions "just works".

schillic:
> Yes, although i would prefer that internally we also use the getter functions

agreed

mforets:
Thanks for your reviews!

Issue #133, opened 2020-01-23
Follow-up of @system macro


mforets:
_Originally posted by @ueliwechsler in https://github.com/JuliaReach/MathematicalSystems.jl/pull/125#issuecomment-575240459_

The following things can be done in a later PR (in approximately increasing order of work needed):

- [ ] add template `f_(x_,u_)` for allowing to write  `@system(x' = f(x,u_12)` (add one line)
- [ ] ensure that `state≠input≠noise`
- [ ] add docstring to all functions
- [ ] make `f_(x_,u_)`  and `f_(x_,w_)` distinguishable 
- [ ] correct handling for arbitrary order of input set definition + unit_tests
- [ ] add `LazySets.Universe(n)` or  `nothing` as default cases for undefined sets (a straightforward  option would be to handle every set as a `Constrained` set with all set as `Universe(n)` by default and if a set definition is provided specify the corresponding set)
- [ ] Make `@system` macro work without paranthesee, e.g. `@system x' = A*x` (adjust argument parsing)
- [ ] Add support for remaining types, e.g.  `polynomial`
- [ ] Add support for `SystemWithOutput`, e.g. `@system(x' = Ax + Bu, y=Bx + Du)`
- [ ] Allow arbitrary input for dynamic equations without `*`, e.g. `@system(x' = Ax + Bu1)`
- [ ] Major revision by using typ-specific "templates" for each `AbstractSystem` type

ueliwechsler:
from #137:
- [ ] allow property initial_state as `state(0) ∈ set` to create `IVP`

ueliwechsler:
from #139 
- [ ] consistent type conversion in  `x' = -x`  and `x' = 2x`

ueliwechsler:
Additionally, we should try to remove the explicit dependency on `MathematicalSystems` in this lines
https://github.com/JuliaReach/MathematicalSystems.jl/blob/641d4f5a067f2bea6fd0dd0de7c2f193a9a8632c/src/macros.jl#L535
https://github.com/JuliaReach/MathematicalSystems.jl/blob/641d4f5a067f2bea6fd0dd0de7c2f193a9a8632c/src/macros.jl#L538

by writing ` push!(params, I($state_dim), :B)` or something similar.

PR #155, opened 2020-02-01
@system: error handling for variable names

ueliwechsler:
Complete an element of checklist in #133.

I am not sure about this point on the checklist. It helps for error handling in general, but in some cases, e.g. if you want to name your state `u` or `w` it would make the macro more verbose (see tests).

### DIFF
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -311,6 +311,11 @@ function parse_system(exprs)
     nsets = length(constraints)
     nsets > 3 && throw(ArgumentError("cannot parse $nsets set constraints"))
 
+    # error handling for variable names
+    (state_var == input_var) && throw(ArgumentError("state and input variable have the same name `$(state_var)`"))
+    (state_var == noise_var) && throw(ArgumentError("state and noise variable have the same name `$(state_var)`"))
+    (input_var == noise_var) && throw(ArgumentError("input and noise variable have the same name `$(input_var)`"))
+
     return dynamic_equation, AT, constraints,
            state_var, input_var, noise_var, dimension
 end
--- a/test/@system.jl
+++ b/test/@system.jl
@@ -60,13 +60,14 @@ end
 end
 
 @testset "@system for linear control continuous systems" begin
-    # if the state should be named `w`
-    @test @system(w' = Aw + Bu) == LinearControlContinuousSystem(A, B)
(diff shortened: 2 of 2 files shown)


mforets:
Sorry, i don't understand the purpose of this change and the purpose of this item list, ` ensure that state≠input≠noise `. Shouldn't we consider `@test @system(x' = Ax + Bw, input:w, noise:w)` a user error? Even if we want to check it, can't we do it without modifying the existing tests? (i mean for example, that `@system(w' = Aw + Bu)` still works? In the case the answer is that it is very hard, do we have a strategy in the longer run?

schillic:
> i mean for example, that `@system(w' = Aw + Bu)` still works?

Yes, that is what I said above, but that is more complicated. The current macro (in `master`) expects a certain variable name for state/input/noise. If you mix those names, the behavior is probably not what you would expect.

> Shouldn't we consider @test `@system(x' = Ax + Bw, input:w, noise:w)` a user error?

Right, and this PR prints an error message instead of doing something unexpected. I find that better than the version in `master`, at least until we support that feature.

ueliwechsler:
> Sorry, i don't understand the purpose of this change and the purpose of this item list

The purpose is to prevent the user from doing things that could have arbitrary behaviour such as having both input and noise defined as `w`, e.g. `@system(x'=Ax+Bw, input:w)`. In this case, it is not well defined what the macro does (from looking at the documentation alone). 

> Even if we want to check it, can't we do it without modifying the existing tests?

I think we should have these checks, it improves the quality, but we could do them more intelligent.
As an alternative method to check the input, we could apply the checks only if the user has defined a new variable name, i,e:
```julia
# we do not check for identical variable names
@system(w'=Aw+Bx)
# we do check for identical variable names because the user changed the bindings
@system(x'=Ax+Bw, input:w)
```

mforets:
Ok, thanks.  I guess what you are saying is to verify if the `state_var`, `input_var` or `noise_var` has been *set*, maybe along these lines:

```julia
    default_state_var = :x
    default_input_var = :u
    default_noise_var = :w

    state_var = nothing
    input_var = nothing
    noise_var = nothing
... 
# if state_var == nothing then it hasn't been set
# do consistency checks
state_var = state_var == nothing ? default_state_var
...
```

ueliwechsler:
I will close this PR, since it is replaced by #164

PR #156, opened 2020-02-01
@system: make macro calls work without parentheses

ueliwechsler:
Complete an element of checklist in #133.

See [here](https://discourse.julialang.org/t/input-parsing-for-macros-without-parantheses/34068/3) , the solution is trivial 🙈

The syntax needs to be 
```julia
@system x' = Ax + Bu  x∈X u∈U
```
without the commas.

I added the corresponding tests

### DIFF
--- a/test/@system.jl
+++ b/test/@system.jl
@@ -41,7 +41,7 @@ f1(x, u, w) = x'*x + u'*u + w'*w
 @testset "@system for continous identity systems" begin
     @test @system(x' = 0, dim: 2) == ContinuousIdentitySystem(2)
     sys = @system x' = 0 dim: 2
-    @test sys  == ContinuousIdentitySystem(2)
+    @test sys == ContinuousIdentitySystem(2)
 
     @test @system(x₁' = 0, dim=2) == ContinuousIdentitySystem(2)
 
--- a/test/@system.jl
+++ b/test/@system.jl
@@ -40,15 +40,22 @@ f1(x, u, w) = x'*x + u'*u + w'*w
 
 @testset "@system for continous identity systems" begin
     @test @system(x' = 0, dim: 2) == ContinuousIdentitySystem(2)
+    sys = @system x' = 0 dim: 2
+    @test sys  == ContinuousIdentitySystem(2)
+
     @test @system(x₁' = 0, dim=2) == ContinuousIdentitySystem(2)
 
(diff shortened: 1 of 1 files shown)


schillic:
I do not like this syntax.

ueliwechsler:
> I do not like this syntax.

Leaving out the commas?

schillic:
Yes, this is not how you write a system definition on paper. I am fine with having this as an option of course. It works already, we can actually not avoid it, and I guess there is no way that `,` works without using `(·)` because this is how Julia parses expressions.
But I suggest to use the syntax `@system(x' = Ax + Bu, x∈X, u∈U)` rather than `@system x' = Ax + Bu x∈X u∈U` when showcasing.

blegat:
I agree with @schillic, this is consistent with
http://www.juliaopt.org/JuMP.jl/v0.20.0/style/#JuMP-macro-syntax-1

mforets:
> See here , the solution is trivial see_no_evil

It is fun that we didn't realize this earlier :)

> this is consistent with http://www.juliaopt.org/JuMP.jl/v0.20.0/style/#JuMP-macro-syntax-1

I agree as well that in `@system` the commas make it easier to parse the expression (for humans). On the other hand this behavior is already available, so i think it is a good idea to add the test proposed in this PR.

test/@system.jl
```diff
@@ -40,15 +40,22 @@ f1(x, u, w) = x'*x + u'*u + w'*w
 
 @testset "@system for continous identity systems" begin
     @test @system(x' = 0, dim: 2) == ContinuousIdentitySystem(2)
+    sys = @system x' = 0 dim: 2
+    @test sys  == ContinuousIdentitySystem(2)
```
```suggestion
    @test sys == ContinuousIdentitySystem(2)
```

test/@system.jl
```diff
@@ -57,6 +64,8 @@ end
 
     @test @system(x' = A*x, x ∈ X) == ConstrainedLinearContinuousSystem(A,X)
     @test @system(x1' = A1x1, x1 ∈ X1) == ConstrainedLinearContinuousSystem(A1,X1)
+    sys = @system x1' =A1x1     x1 ∈ X1
+    @test sys  == ConstrainedLinearContinuousSystem(A1,X1)
```
```suggestion
    @test sys == ConstrainedLinearContinuousSystem(A1,X1)
```

PR #158, opened 2020-02-07
Add traits for system types

### LINKED ISSUE #130: Extend traits to system types
See the discussion [here](https://github.com/JuliaReach/MathematicalSystems.jl/pull/122#discussion_r360704941).

ueliwechsler:
Closes #130

### DIFF
--- a/src/abstract.jl
+++ b/src/abstract.jl
@@ -60,10 +60,10 @@ linear in ``(x, u)`` for each ``t ∈ \\mathbb{R}``. On the other hand,
 ``x' = f(t, x, u) = A x + B u + c`` is affine but not linear, since it is not
 linear in ``(x, u)``.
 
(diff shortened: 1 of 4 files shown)


ueliwechsler:
Do I also need to add new method descriptions in `abstract.jl`?

schillic:
> Do I also need to add new method descriptions in `abstract.jl`?

Instead of adding new methods, maybe just add a sentence to the existing methods:

    The result of this function only depends on the system type and can also be applied to `typeof(s)`.

Some of the docs already contain a sentence like this (see example below) but they are not always written in the same way. I would use the same sentence everywhere.

https://github.com/JuliaReach/MathematicalSystems.jl/blob/3e81f81882e3e4af821d03a546547b38e775d5a0/src/abstract.jl#L63-L66

PR #164, opened 2020-02-11
Error handling for repeated variable names

mforets:
From discussion on gitter, consider this PR in place of  https://github.com/JuliaReach/MathematicalSystems.jl/pull/155.

### DIFF
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -251,7 +251,6 @@ function parse_system(exprs)
     constraints = Vector{Expr}()
 
     # define defaults for state, noise and input symbol and default dimension
-    default_state_var = :x
     default_input_var = :u
     default_noise_var = :w
     state_var = nothing
@@ -325,21 +324,18 @@ function parse_system(exprs)
     nsets > 3 && throw(ArgumentError("cannot parse $nsets set constraints"))
 
     # error handling for variable names
-    got_state_var = state_var != nothing
+    state_var == nothing && throw(ArgumentError("the state variable was not found"))
     got_input_var = input_var != nothing
     got_noise_var = noise_var != nothing
-    if got_state_var && got_input_var && (state_var == input_var)
+    if got_input_var && (state_var == input_var)
          throw(ArgumentError("state and input variables have the same name `$(state_var)`"))
-    elseif got_state_var && got_noise_var && (state_var == noise_var)
+    elseif got_noise_var && (state_var == noise_var)
          throw(ArgumentError("state and noise variables have the same name `$(state_var)`"))
     elseif got_input_var && got_noise_var && (input_var == noise_var)
          throw(ArgumentError("input and noise variables have the same name `$(input_var)`"))
     end
 
     # assign default values
-    if !got_state_var
-        state_var = default_state_var
-    end
     if !got_input_var
         input_var = default_input_var
     end
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -324,7 +324,7 @@ function parse_system(exprs)
          throw(ArgumentError("state and input variables have the same name `$(state_var)`"))
     elseif got_state_var && got_noise_var && (state_var == noise_var)
          throw(ArgumentError("state and noise variables have the same name `$(state_var)`"))
-    elseif (got_input_var && got_noise_var) && (input_var == noise_var)
+    elseif got_input_var && got_noise_var && (input_var == noise_var)
          throw(ArgumentError("input and noise variables have the same name `$(input_var)`"))
     end
 
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -320,7 +320,7 @@ function parse_system(exprs)
     got_state_var = state_var != nothing
     got_input_var = input_var != nothing
     got_noise_var = noise_var != nothing
-    if (got_state_var && got_input_var) && (state_var == input_var)
+    if got_state_var && got_input_var && (state_var == input_var)
          throw(ArgumentError("state and input variables have the same name `$(state_var)`"))
     elseif (got_state_var && got_noise_var) && (state_var == noise_var)
          throw(ArgumentError("state and noise variables have the same name `$(state_var)`"))
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -322,7 +322,7 @@ function parse_system(exprs)
     got_noise_var = noise_var != nothing
     if got_state_var && got_input_var && (state_var == input_var)
          throw(ArgumentError("state and input variables have the same name `$(state_var)`"))
-    elseif (got_state_var && got_noise_var) && (state_var == noise_var)
+    elseif got_state_var && got_noise_var && (state_var == noise_var)
          throw(ArgumentError("state and noise variables have the same name `$(state_var)`"))
     elseif (got_input_var && got_noise_var) && (input_var == noise_var)
          throw(ArgumentError("input and noise variables have the same name `$(input_var)`"))
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -252,9 +252,12 @@ function parse_system(exprs)
     constraints = Vector{Expr}()
 
     # define defaults for state, noise and input symbol and default dimension
-    state_var = :x
-    input_var = :u
-    noise_var = :w
+    default_state_var = :x
+    default_input_var = :u
+    default_noise_var = :w
+    state_var = nothing
+    input_var = nothing
+    noise_var = nothing
     dimension = nothing
 
     # main loop to parse the subexpressions in exprs
@@ -313,8 +316,29 @@ function parse_system(exprs)
     nsets = length(constraints)
     nsets > 3 && throw(ArgumentError("cannot parse $nsets set constraints"))
 
-    return dynamic_equation, AT, constraints,
-           state_var, input_var, noise_var, dimension
+    # error handling for variable names
+    got_state_var = state_var != nothing
+    got_input_var = input_var != nothing
+    got_noise_var = noise_var != nothing
+    if (got_state_var && got_input_var) && (state_var == input_var)
+         throw(ArgumentError("state and input variables have the same name `$(state_var)`"))
+    elseif (got_state_var && got_noise_var) && (state_var == noise_var)
+         throw(ArgumentError("state and noise variables have the same name `$(state_var)`"))
+    elseif (got_input_var && got_noise_var) && (input_var == noise_var)
+         throw(ArgumentError("input and noise variables have the same name `$(input_var)`"))
+    end
+
+    # assign default values
+    if !got_state_var
+        state_var = default_state_var
+    end
+    if !got_input_var
+        input_var = default_input_var
+    end
+    if !got_noise_var
+        noise_var = default_noise_var
+    end
+    return dynamic_equation, AT, constraints, state_var, input_var, noise_var, dimension
 end
 
 # extract the field and value parameter from the dynamic equation `equation`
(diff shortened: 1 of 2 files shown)


src/macros.jl
```diff
@@ -313,8 +316,29 @@ function parse_system(exprs)
     nsets = length(constraints)
     nsets > 3 && throw(ArgumentError("cannot parse $nsets set constraints"))
 
-    return dynamic_equation, AT, constraints,
-           state_var, input_var, noise_var, dimension
+    # error handling for variable names
+    got_state_var = state_var != nothing
+    got_input_var = input_var != nothing
+    got_noise_var = noise_var != nothing
+    if (got_state_var && got_input_var) && (state_var == input_var)
```
```suggestion
    if got_state_var && got_input_var && (state_var == input_var)
```

src/macros.jl
```diff
@@ -313,8 +316,29 @@ function parse_system(exprs)
...
+         throw(ArgumentError("state and input variables have the same name `$(state_var)`"))
+    elseif (got_state_var && got_noise_var) && (state_var == noise_var)
```
```suggestion
    elseif got_state_var && got_noise_var && (state_var == noise_var)
```

src/macros.jl
```diff
@@ -313,8 +316,29 @@ function parse_system(exprs)
...
+         throw(ArgumentError("state and noise variables have the same name `$(state_var)`"))
+    elseif (got_input_var && got_noise_var) && (input_var == noise_var)
```
```suggestion
    elseif got_input_var && got_noise_var && (input_var == noise_var)
```

src/macros.jl
```diff
@@ -313,8 +316,29 @@ function parse_system(exprs)
...
+    # assign default values
+    if !got_state_var
+        state_var = default_state_var
+    end
+    if !got_input_var
+        input_var = default_input_var
+    end
+    if !got_noise_var
+        noise_var = default_noise_var
+    end
```
schillic: I think you want to make sure that the default is different from the assigned names. Otherwise you run into the same problem as before.

mforets: Not sure, why? If a default is the same as an assigned name, that is not a problem.

schillic: You can reach a state below this block where `state_var == input_var`. Why is this not a problem?

ueliwechsler: We only want to throw an error if we changed the default values which is checked e.g by `got_state_var`

schillic: The code *only* checks for a clash between defined variable names. Example: `state_var: y, input_var: y`.
But the code *does not* check for a clash with the names that will be used later (after assigning the default values). Example: `state_var: nothing, input_var: x`. After assigning the default, you get `state_var: x, input_var: x`. The fact that this does not crash in practice is that `got_state_var` is invariably `true` because that is read from the left-hand side of the equation. So I suggest to remove this variable as it is confusing.
Here is an example that shows what I originally meant (using `input: u` and `noise: nothing`):
```julia
julia> @system(x⁺ = Ax + Bu + Du, input: u)
ERROR: ArgumentError: the entry (:A, :B, :B) does not match a `MathematicalSystems.jl` structure
```

mforets: > So I suggest to remove this variable as it is confusing.

Alright, i have removed that variable (and added an independent check that it has been assigned).

schillic: My example above still does not work.

mforets: I got your point, see the last commit where i swapped the argument checks and simplified the logic. OTOH, the assigned names for the input and noise in `@system(x⁺ = Ax + Bu + Du, input: u)` are `u` and `w` inside this function, because at that site it doesn't know what type of system this is. In other words, i don't think that we can handle this case here without bigger changes.

ueliwechsler: Ì think we could count how many time the `state`, `input` or `noise` was "found" in `extract_sum` to address this problem without bigger changes.

ueliwechsler: Something like this should work
```julia
function extract_sum(summands, state::Symbol, input::Symbol, noise::Symbol)
    params = Tuple{Any,Symbol}[]
    state_dim = 1
    got_state_dim = false

    num_state_assignements = 0
    num_input_assignements = 0
    num_noise_assignements = 0
    num_const_assignements = 0

    for summand in summands
        if @capture(summand, array_ * var_)
            if state == var
                push!(params, (array, :A))
                # obtain "state_dim" for later using in IdentityMultiple
                state_dim =  Expr(:call, :size, :($array), 1)
                got_state_dim = true
                num_state_assignements += 1

            elseif input == var
                push!(params, (array, :B))
                num_input_assignements += 1

            elseif noise == var
                push!(params, (array, :D))
                num_noise_assignements += 1

            else
                throw(ArgumentError("in the dynamic equation, the expression "*
                "$summand does not contain the state $state, the input $input "*
                "or the noise term $noise"))
            end
        elseif @capture(summand, array_)
            if got_state_dim
                identity = :(I($state_dim))
            else
                identity = 1.0
            end
            # if array == variable: field value equals identity
            if state == array
                push!(params, (identity, :A))
                num_state_assignements += 1
            elseif input == array
                push!(params, (identity, :B))
                num_input_assignements += 1
            elseif noise == array
                push!(params, (identity, :D))
                num_noise_assignements += 1
            else
                push!(params, (array, :c))
                num_const_assignements += 1
            end
        end
    end
    num_const_assignements > 1 && throw(ArgumentError("there is more than one constant term"))
    num_state_assignements > 1 && throw(ArgumentError("there is more than one state term $state"))
    num_input_assignements > 1 && throw(ArgumentError("there is more than one input term $input"))
    num_noise_assignements > 1 && throw(ArgumentError("there is more than one noise term $noise"))

    return params
end
```

schillic: > i don't think that we can handle this case here without bigger changes.

I think it is pretty simple: just compare the default values to the other values, or simpler: compare the final values *after* assigning the default values again.

mforets: for `@system(x⁺ = Ax + Bu + Du, input: u)`, [parse_system](https://github.com/JuliaReach/MathematicalSystems.jl/blob/a1d94d6658d6e7e6983e9fd5ef7783c20fb72b5f/src/macros.jl#L246) returns with `state_var -> :x`, `input_var -> :u` and `noise_var -> :w`, so there is no collision after assigning default values.

mforets: note that `parse_system` only changes defaults if it captures the statement `@capture(ex, (input = u_) | (u_ = input))`. There is also some logic to capture the [input](https://github.com/JuliaReach/MathematicalSystems.jl/blob/a1d94d6658d6e7e6983e9fd5ef7783c20fb72b5f/src/macros.jl#L273) from the form of the system, but it is not done for the noise.

schillic: > for `@system(x⁺ = Ax + Bu + Du, input: u)`, [parse_system](https://github.com/JuliaReach/MathematicalSystems.jl/blob/a1d94d6658d6e7e6983e9fd5ef7783c20fb72b5f/src/macros.jl#L246) returns with `state_var -> :x`, `input_var -> :u` and `noise_var -> :w`, so there is no collision after assigning default values.

I see. Still, from a programming point of view, I find it uncomfortable to assign a default value without checking compatibility with the other values that already exist. Here is another example where I added
```julia
    show(state_var)
    println()
    show(input_var)
    println()
    show(noise_var)
    println()
```
before and after the assignment with default values:
```julia
julia> @system(x⁺ = Ax + Bu + Du, noise: u)
# before assignment
:x
nothing
:u
# after assignment
:x
:u
:u
```

ueliwechsler: I get your point. But we "allow for conflicting variable names" in order to keep the interface as minimal as possible, as long as a variable is not used twice in the right-hand side.
`@system(x' = Ax + Bw, input:w)` would not work if we have more rigorous checks.

To see how we could throw an error for `@system(x⁺ = Ax + Bu + Du`
see https://github.com/JuliaReach/MathematicalSystems.jl/pull/164#discussion_r380758870

mforets: We can also make the default assignment in another function, i mean after `parse_system`, i.e. such that `parse_system` returns `nothing` if the variable hasn't been identifies at that point of the program.

schillic: >  `@system(x' = Ax + Bw, input:w)` would not work if we have more rigorous checks.

Only if you assign the values blindly. Of course you should not assign `w` as noise variable if there is no noise.

> But we "allow for conflicting variable names" in order to keep the interface as minimal as possible, as long as a variable is not used twice in the right-hand side.

But in my examples the variables were used twice.

> We can also make the default assignment in another function, i mean after `parse_system`, i.e. such that `parse_system` returns `nothing` if the variable hasn't been identified at that point of the program.

Okay, but I do not see how that solves the problem cases I mentioned.

ueliwechsler: > But in my examples the variables were used twice.

For this to work, we need to add https://github.com/JuliaReach/MathematicalSystems.jl/pull/164#discussion_r380758870 or something similar

schillic: Okay, I give up since nobody seems to care that two variables have the same value, which I thought was the whole point of this PR.

ueliwechsler:
After thinking everything through more again more carefully, I think you are right.

In order for me to understand what is going on, I like to write everything I know down
(sorry for the long comment)

As far as i know, this PR should give reasonable error messages for cases where
a specified `input` or `noise` is added which does collide with another  
```julia
A = B = D = rand(2,2)
@system(a⁺= Aa + Bw, input:a)
ERROR: ArgumentError: state and input variables have the same name `a`
# instead of the old error message
ERROR: ArgumentError: the entry (:A, :D) does not match a `MathematicalSystems.jl` structure
@system(x⁺= Ax + Bx, input:x)
ERROR: ArgumentError: state and input variables have the same name `x`
@system(x⁺= Ax + Bu + Bw, input:u, noise:u)
ERROR: ArgumentError: input and noise variables have the same name `u`
```
without throwing exceptions where the input is "overwritten" with the noise variable 
(and the noise not defined) or vice versa where the input is overwritten 
```julia
# we can use 
@system(x⁺= Ax + Bw, input:w) #creates a LinearControlDiscreteSystem system
system(x⁺= Ax + Bu, noise:u) # 
# instead of 
@system(x⁺= Ax + Bw, input:w, noisestr:evry_but_w)
````
in the `@system(x⁺= Ax + Bw, input:w)`, the input and the noise variable are both `:w` but 
since we do not "use" the noise variable we accept that "inconsistency" for shortness of the
macro.

However, this cases are not covered so far in this PR (but we also should add it)
```julia
@system(x⁺= Ax + Bw + Dw, input:w)
@system(x⁺= Ax + Bu + Bu,  noise:u)
@system(x⁺= Ax + Bx)
@system(x⁺= Ax + c + c)
# etc
```
which gives a non-sensical error messsage. (for which my aforementioned fix 
should help).

Hoooowever, going through this logic it should also give the right error for
```julia
@system(x⁺= Ax + Bu + Bw,  noise:u)
ERROR: ArgumentError: the entry (:A, :B, :c) does not match a `MathematicalSystems.jl` structure
@system(x⁺= Ax + Bu + Bw,  input:w)
ERROR: ArgumentError: the entry (:A, :c, :B) does not match a `MathematicalSystems.jl` structure
```
which it does not, and also will not get caught by my solutions.

Sooo, too sum up, I am bit confused. I think we could do it better, you are right, 
but I am not sure how. Maybe throwing the error at a later stage, where we no the
exact system types and all the variables.

mforets:
> Maybe throwing the error at a later stage, where we no the
exact system types and all the variables.

Yes, that is what i thought (but i haven't tried it).

:+1: i moved your comment to the follow up issue, https://github.com/JuliaReach/MathematicalSystems.jl/issues/181

ueliwechsler:
Thanks! But we still have to catch the error from `@system(x⁺= Ax + Bu + Bw,  noise:u)` and `@system(x⁺= Ax + Bu + Bw,  input:w)` I think this one could be caught after here after this line https://github.com/JuliaReach/MathematicalSystems.jl/blob/a1d94d6658d6e7e6983e9fd5ef7783c20fb72b5f/src/macros.jl#L807... it is getting quite ugly if we have 3 separate places for tests.

Issue #140, opened 2020-01-28
Parse scalar affine continuous system


mforets:
Similar to `@system(x' = 2x)`, i think we should be able to parse:

```julia
julia> @system(x' = 2x + 1)
ERROR: MethodError: no method matching AffineContinuousSystem(::Int64, ::Int64)
Stacktrace:
 [1] top-level scope at REPL[9]:1
```

ueliwechsler:
Same goes for `@system(x⁺ = 0.5x + u)`
```julia
MethodError: no method matching LinearControlDiscreteSystem(::Float64, ::IdentityMultiple{Float64})
Closest candidates are:
  LinearControlDiscreteSystem(!Matched::MTA, ::MTB) where {T, MTA<:AbstractArray{T,2}, MTB<:AbstractArray{T,2}} at C:\Users\wueli\.julia\dev\MathematicalSystems\src\systems.jl:222
in top-level scope at ifmpc.jl:10
```

ueliwechsler:
I am not sure how to handle `@system(x⁺ = 0.5x + u)`  
```julia
MethodError: no method matching LinearControlDiscreteSystem(::Float64, ::IdentityMultiple{Float64})
```
Either we need to convert the `Float64` argument to a `Matrix{Float64}` or the `IdentityMultiple` to a scalar (or change the signature of the constructors). 
The `@system` macro needs to check if the dimension of `A=0.5` is `1` (then we  need `B=1`) or bigger than `1` (then we need `B=MultipleIdentity(state_dim)`).
So far, I couldn't make the dynamic dimension check work within the macro.

schillic:
> Either we need to convert the Float64 argument to a Matrix{Float64} or the IdentityMultiple to a scalar (or change the signature of the constructors).

I think the behavior should be consistent with `@system(x⁺ = x)` and `@system(x⁺ = x + u)`.

ueliwechsler:
That means it should create a ` LinearControlDiscreteSystem(::Float64, ::Float64)`.

schillic:
Or we change the behavior of the other cases. Either way is fine with me but other people may have an opinion.

ueliwechsler:
I did some thinking 😅
A proposal would be to convert every input into a `matrix` or `vector` see #175 as a possible solution.

PR #167, opened 2020-02-16
Add constructor for scalar arguments

### LINKED ISSUE #141: MethodError for linear system x' = ax
```julia
julia> a = 2.0
2.0

julia> @system(x' = ax)
ERROR: MethodError: no method matching LinearContinuousSystem(::Float64)
Closest candidates are:
  LinearContinuousSystem(::MT<:AbstractArray{T,2}) where {T, MT<:AbstractArray{T,2}} at /home/mforets/.julia/dev/MathematicalSystems/src/systems.jl:114
Stacktrace:
 [1] top-level scope at REPL[11]:1
```

ueliwechsler:
Closes #141 and first example in #140

### DIFF
--- a/src/systems.jl
+++ b/src/systems.jl
@@ -125,6 +125,11 @@ for (Z, AZ) in ((:LinearContinuousSystem, :AbstractContinuousSystem),
                 return new{T, MT}(A)
             end
         end
+        function $(Z)(A::Real)
+             MA = Matrix{typeof(A)}(undef, 1, 1); MA[1, 1] = A
+             return $(Z)(MA)
+        end
+
         statedim(s::$Z) = size(s.A, 1)
         inputdim(::$Z) = 0
         noisedim(::$Z) = 0
@@ -183,6 +188,12 @@ for (Z, AZ) in ((:AffineContinuousSystem, :AbstractContinuousSystem),
                 return new{T, MT, VT}(A, c)
             end
         end
+        function $(Z)(A::Real, c::Real)
+             MA = Matrix{typeof(A)}(undef, 1, 1); MA[1, 1] = A
+             Vc = Vector{typeof(c)}(undef, 1); Vc[1] = c
+             return $(Z)(MA, Vc)
+        end
+
         statedim(s::$Z) = length(s.c)
         inputdim(::$Z) = 0
         noisedim(::$Z) = 0
@@ -244,6 +255,12 @@ for (Z, AZ) in ((:LinearControlContinuousSystem, :AbstractContinuousSystem),
                 return new{T, MTA, MTB}(A, B)
             end
         end
+        function $(Z)(A::Real, B::Real)
+             MA = Matrix{typeof(A)}(undef, 1, 1); MA[1, 1] = A
+             MB = Matrix{typeof(B)}(undef, 1, 1); MB[1, 1] = B
+             return $(Z)(MA, MB)
+        end
+
         statedim(s::$Z) = size(s.A, 1)
         inputdim(s::$Z) = size(s.B, 2)
         noisedim(::$Z) = 0
@@ -305,6 +322,11 @@ for (Z, AZ) in ((:ConstrainedLinearContinuousSystem, :AbstractContinuousSystem),
                 return new{T, MT, ST}(A, X)
             end
         end
+        function $(Z)(A::Real, X)
+             MA = Matrix{typeof(A)}(undef, 1, 1); MA[1, 1] = A
+             return $(Z)(MA, X)
+        end
+
         statedim(s::$Z) = size(s.A, 1)
         inputdim(::$Z) = 0
         noisedim(::$Z) = 0
@@ -367,6 +389,12 @@ for (Z, AZ) in ((:ConstrainedAffineContinuousSystem, :AbstractContinuousSystem),
                 return new{T, MT, VT, ST}(A, c, X)
             end
         end
+        function $(Z)(A::Real, c::Real, X)
+             MA = Matrix{typeof(A)}(undef, 1, 1); MA[1, 1] = A
+             Vc = Vector{typeof(c)}(undef, 1); Vc[1] = c
+             return $(Z)(MA, Vc, X)
+        end
+
         statedim(s::$Z) = length(s.c)
         inputdim(::$Z) = 0
         noisedim(::$Z) = 0
@@ -434,6 +462,13 @@ for (Z, AZ) in ((:ConstrainedAffineControlContinuousSystem, :AbstractContinuousS
                 return new{T, MTA, MTB, VT, ST, UT}(A, B, c, X, U)
             end
         end
+        function $(Z)(A::Real, B::Real, c::Real, X, U)
+             MA = Matrix{typeof(A)}(undef, 1, 1); MA[1, 1] = A
+             MB = Matrix{typeof(B)}(undef, 1, 1); MB[1, 1] = B
+             Vc = Vector{typeof(c)}(undef, 1); Vc[1] = c
+             return $(Z)(MA, MB, Vc, X, U)
+        end
+
         statedim(s::$Z) = length(s.c)
         inputdim(s::$Z) = size(s.B, 2)
         noisedim(::$Z) = 0
@@ -508,6 +543,12 @@ for (Z, AZ) in ((:ConstrainedLinearControlContinuousSystem, :AbstractContinuousS
                 return new{T, MTA, MTB, ST, UT}(A, B, X, U)
             end
         end
+        function $(Z)(A::Real, B::Real, X, U)
+             MA = Matrix{typeof(A)}(undef, 1, 1); MA[1, 1] = A
+             MB = Matrix{typeof(B)}(undef, 1, 1); MB[1, 1] = B
+             return $(Z)(MA, MB, X, U)
+        end
+
         statedim(s::$Z) = size(s.A, 1)
         inputdim(s::$Z) = size(s.B, 2)
         noisedim(::$Z) = 0
@@ -574,6 +615,13 @@ for (Z, AZ) in ((:LinearAlgebraicContinuousSystem, :AbstractContinuousSystem),
                 return new{T, MTA, MTE}(A, E)
             end
         end
+        function $(Z)(A::Real, E::Real)
+             MA = Matrix{typeof(A)}(undef, 1, 1); MA[1, 1] = A
+             ME = Matrix{typeof(E)}(undef, 1, 1); ME[1, 1] = E
+             return $(Z)(MA, ME)
+        end
+
+
         statedim(s::$Z) = size(s.A, 1)
         inputdim(::$Z) = 0
         noisedim(::$Z) = 0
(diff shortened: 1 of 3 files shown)


src/systems.jl
```diff
@@ -183,6 +188,12 @@ for (Z, AZ) in ((:AffineContinuousSystem, :AbstractContinuousSystem),
                 return new{T, MT, VT}(A, c)
             end
         end
+        function $(Z)(A::Real, c::Real)
+             MA = Matrix{typeof(A)}(undef, 1, 1); MA[1, 1] = A
+             Vc = Vector{typeof(c)}(undef, 1); Vc[1] = c
```
schillic:
```suggestion
             Vc = [c]
```
I guess we could use Julia's convenience constructor for `Vector`s here as well, but we can also keep it as is. But then I suggest to write a helper function again.

```julia
julia> c = 1.0f32
1.0f32

julia> [c]
1-element Array{Float32,1}:
 1.0e32

julia> Vc = Vector{typeof(c)}(undef, 1); Vc[1] = c; Vc
1-element Array{Float32,1}:
 1.0e32

julia> @btime [$c]
  28.561 ns (1 allocation: 96 bytes)
1-element Array{Float32,1}:
 1.0e32

julia> @btime (Vc = Vector{typeof($c)}(undef, 1); Vc[1] = $c; Vc)
  30.418 ns (1 allocation: 96 bytes)
1-element Array{Float32,1}:
 1.0e32
```

src/systems.jl
```diff
@@ -125,6 +125,11 @@ for (Z, AZ) in ((:LinearContinuousSystem, :AbstractContinuousSystem),
                 return new{T, MT}(A)
             end
         end
+        function $(Z)(A::Real)
+             MA = Matrix{typeof(A)}(undef, 1, 1); MA[1, 1] = A
```
schillic: Since this pattern comes up so often here, how about defining a helper function? This would also be more readable.
"Find and replace" will be your friend here.

test/continuous.jl
```diff
@@ -62,6 +72,11 @@ end
             @test !isnoisy(s) && iscontrolled(s) && !isconstrained(s)
         end
     end
+    # Scalar System
+    a = 1.; b=2.
```
schillic:
```suggestion
    a = 1.; b = 2.
```

test/continuous.jl
```diff
@@ -155,6 +185,11 @@ end
             @test !isnoisy(s) && !iscontrolled(s) && !isconstrained(s)
         end
     end
+    # Scalar System
+    a = 1.; e = 2.;
```
schillic:
```suggestion
    a = 1.; e = 2.
```

test/continuous.jl
```diff
@@ -155,6 +185,11 @@ end
...
+    A = [a][:,:]; E = [e][:,:];
```
schillic:
```suggestion
    A = [a][:,:]; E = [e][:,:]
```

test/continuous.jl
```diff
@@ -170,6 +205,11 @@ end
         @test islinear(s) && isaffine(s) && !ispolynomial(s)
         @test !isnoisy(s) && !iscontrolled(s) && isconstrained(s)
     end
+    # Scalar System
+    a = 1.; e = 2.; X = 1
+    A = [a][:,:]; E = [e][:,:];
```
schillic:
```suggestion
    A = [a][:,:]; E = [e][:,:]
```

test/continuous.jl
```diff
@@ -311,6 +354,11 @@ end
         @test islinear(s) && isaffine(s) && !ispolynomial(s)
         @test isnoisy(s) && !iscontrolled(s) && isconstrained(s)
     end
+    # Scalar System
+    a = 1.; d = 3.; X = 1; W = 3
+    A = [a][:,:]; D = [d][:,:];
```
schillic:
```suggestion
    A = [a][:,:]; D = [d][:,:]
```

test/continuous.jl
```diff
@@ -334,6 +382,11 @@ end
         @test islinear(s) && isaffine(s) && !ispolynomial(s)
         @test isnoisy(s) && iscontrolled(s) && isconstrained(s)
     end
+    # Scalar System
+    a = 1.; b = 2.; d = 3.; X = 1; U = 2; W = 3
+    A = [a][:,:]; B = [b][:,:]; D = [d][:,:];
```
schillic:
```suggestion
    A = [a][:,:]; B = [b][:,:]; D = [d][:,:]
```

test/continuous.jl
```diff
@@ -359,6 +412,11 @@ end
         @test isaffine(s) && !islinear(s) && !ispolynomial(s)
         @test isnoisy(s) && iscontrolled(s) && isconstrained(s)
     end
+    # Scalar System
+    a = 1.; b = 2.; c = 0.1; d = 3.; X = 1; U = 2; W = 3
+    A = [a][:,:]; B = [b][:,:]; C = [c]; D = [d][:,:];
```
schillic:
```suggestion
    A = [a][:,:]; B = [b][:,:]; C = [c]; D = [d][:,:]
```

test/discrete.jl
```diff
@@ -24,6 +24,7 @@ end
             @test !isnoisy(s) && !iscontrolled(s) && isconstrained(s)
         end
     end
+
```
schillic:
```suggestion
```

test/discrete.jl
```diff
@@ -63,6 +74,11 @@ end
             @test !isnoisy(s) && iscontrolled(s) && !isconstrained(s)
         end
     end
+    # Scalar System
+    a = 1.; b=2.
```
schillic:
```suggestion
    a = 1.; b = 2.
```

test/discrete.jl
```diff
@@ -109,6 +135,11 @@ end
         @test islinear(s) && isaffine(s) && !ispolynomial(s)
         @test !isnoisy(s) && iscontrolled(s) && isconstrained(s)
     end
+    # Scalar System
+    a = 1.; b = 2.; X = 1; U = 2
+    A = [a][:,:]; B = [b][:,:];
```
schillic:
```suggestion
    A = [a][:,:]; B = [b][:,:]
```

test/discrete.jl
```diff
@@ -122,6 +153,11 @@ end
             @test !isnoisy(s) && !iscontrolled(s) && !isconstrained(s)
         end
     end
+    # Scalar System
+    a = 1.; e = 2.;
+    A = [a][:,:]; E = [e][:,:];
```
schillic:
```suggestion
    A = [a][:,:]; E = [e][:,:]
```

test/discrete.jl
```diff
@@ -137,6 +173,11 @@ end
         @test islinear(s) && isaffine(s) && !ispolynomial(s)
         @test !isnoisy(s) && !iscontrolled(s) && isconstrained(s)
     end
+    # Scalar System
+    a = 1.; e = 2.; X = 1
+    A = [a][:,:]; E = [e][:,:];
```
schillic:
```suggestion
    A = [a][:,:]; E = [e][:,:]
```

test/discrete.jl
```diff
@@ -195,6 +240,11 @@ end
         @test islinear(s) && isaffine(s) && !ispolynomial(s)
         @test isnoisy(s) && !iscontrolled(s) && isconstrained(s)
     end
+    # Scalar System
+    a = 1.; d = 3.; X = 1; W = 3
+    A = [a][:,:]; D = [d][:,:];
```
schillic:
```suggestion
    A = [a][:,:]; D = [d][:,:]
```

test/discrete.jl
```diff
@@ -218,6 +268,11 @@ end
         @test islinear(s) && isaffine(s) && !ispolynomial(s)
         @test isnoisy(s) && iscontrolled(s) && isconstrained(s)
     end
+    # Scalar System
+    a = 1.; b = 2.; d = 3.; X = 1; U = 2; W = 3
+    A = [a][:,:]; B = [b][:,:]; D = [d][:,:];
```
schillic:
```suggestion
    A = [a][:,:]; B = [b][:,:]; D = [d][:,:]
```

test/discrete.jl
```diff
@@ -243,6 +298,11 @@ end
         @test isaffine(s) && !islinear(s) && !ispolynomial(s)
         @test isnoisy(s) && iscontrolled(s) && isconstrained(s)
     end
+    # Scalar System
+    a = 1.; b = 2.; c = 0.1; d = 3.; X = 1; U = 2; W = 3
+    A = [a][:,:]; B = [b][:,:]; C = [c]; D = [d][:,:];
```
schillic:
```suggestion
    A = [a][:,:]; B = [b][:,:]; C = [c]; D = [d][:,:]
```

src/systems.jl
mforets: FWIW, `hcat(A)` for scalar `A` works too (def. [here](https://github.com/JuliaLang/julia/blob/c6da87ff4bc7a855e217856757ad3413cf6d1f79/base/abstractarray.jl#L1280)); i wouldn't use a helper function for that.

src/systems.jl
mforets: Here we can use `vcat(c)` :)

schillic: > Here we can use `vcat(c)` :)

Is that better than `[c]`?

ueliwechsler: It is identical.

src/systems.jl
ueliwechsler: So I will replace all the verbosity with `hcat` and `vcat`, okay?

src/systems.jl
mforets: Here i would use either `[c]` or `vcat(c)`, i don't care which. The performance of the 3 methods considered is the same (try wrapping each into a function for `@btime`).

src/systems.jl
schillic: Sorry for not seeing `hcat` in the issue already :eyes:

src/systems.jl
ueliwechsler: If you do `@code_native` it give you exactly the same assembly instruction

src/systems.jl
ueliwechsler: No worries :) I am glad to know this feature now, I was not aware of that.

mforets: Sorry to reopen the conversation, but wouldn't this be a good use case for the identity multiple? Also, that's the current behavior for:

```julia
julia> @system(x' = 2x)
LinearContinuousSystem{Float64,IdentityMultiple{Float64}}([2.0])
```

schillic: I think `@system` should produce a standard `Matrix` that is compatible with most other code. `IdentityMultiple` can cause too many problems (take #171 from yesterday). You can always create the matrix of your choice if you want to.

ueliwechsler:
I did not think about complex number in this context, but I don't see any problem in using `Number`.

src/systems.jl
mforets: Ok, ill defer this discussion a separate issue.

mforets:
True, i think `Number` is better.

PR #169, opened 2020-02-16
Address type stability of x` = a*x 

### LINKED ISSUE #139: Numeric type not preserved in @system
With:

```julia
julia> @system(x' = 2x)
LinearContinuousSystem{Float64,IdentityMultiple{Float64}}([2.0])
```
I expected that the numeric type of this system is an integer. Actually, I like to keep it as is, but we should stress this point in the docs (so i leave this issue open, to consider the documentation update).

ueliwechsler:
Closes #139 and add some minor docstring improvements

### DIFF
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -323,8 +323,8 @@ end
 Extract the value and field parameter from the dynamic equation `equation` according
 to the variable `state`, `input` and `noise`.
 
-For the right-hand side of the dynamic equation, it has returns a vector of tuples
-which consiste one or more of the elements
+For the right-hand side of the dynamic equation, this function returns a vector of tuples
+containing some elements from the list
 - `(:A_user, :A)`
 - `(:B_user, :B)`
 - `(:c_user, :c)`
@@ -333,9 +333,9 @@ which consiste one or more of the elements
 - (:statedim_user :statedim)`
 - (:inputdim_user :inputdim)`
 - (:noisedim_user :noisedim)`
-and for the left-hand side, it returns either a empty vector `Any[]` or
+and for the left-hand side, it returns either an empty vector `Any[]` or
 `[(:E_user, :E)]` where the first argument of the tuple corresponds to the value
-and the second to the field parameter.
+and the second argument of the tuple corresponds to the field parameter.
 
 ### Input
 
@@ -348,7 +348,7 @@ and the second to the field parameter.
 
 ### Output
 
-Two arrays of tuples containing the field and value parameter for the right-hand
+Two arrays of tuples containing the value and field parameters for the right-hand
 and left-hand side of the dynamic equation `equation`.
 """
 function extract_dyn_equation_parameters(equation, state, input, noise, dim, AT)
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -330,21 +330,21 @@ containing some elements from the list
 - `(:c_user, :c)`
 - `(:D_user, :D)`
 - `(:f_user, :f)`
-- (:statedim_user :statedim)`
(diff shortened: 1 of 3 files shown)


src/macros.jl
```diff
@@ -317,9 +317,40 @@ function parse_system(exprs)
...
-# extract the field and value parameter from the dynamic equation `equation`
-# in the form `rhs = [(:A_user, :A), (:B_user, :B), ....]` and
-# `lhs = [(:E_user, :E)]` or `lhs = Any[]`
+"""
+    extract_dyn_equation_parameters(equation, state, input, noise, dim, AT)
+
+Extract the value and field parameter from the dynamic equation `equation` according
+to the variable `state`, `input` and `noise`.
+
+For the right-hand side of the dynamic equation, it has returns a vector of tuples
```
```suggestion
For the right-hand side of the dynamic equation, this function returns a vector of tuples
```

src/macros.jl
```diff
@@ -317,9 +317,40 @@ function parse_system(exprs)
...
+which consiste one or more of the elements
```
```suggestion
containing some elements from the list
```

src/macros.jl
```diff
@@ -317,9 +317,40 @@ function parse_system(exprs)
...
+- `(:A_user, :A)`
+- `(:B_user, :B)`
+- `(:c_user, :c)`
+- `(:D_user, :D)`
+- `(:f_user, :f)`
+- (:statedim_user :statedim)`
+- (:inputdim_user :inputdim)`
+- (:noisedim_user :noisedim)`
```
schillic: leading ` missing

src/macros.jl
```diff
@@ -317,9 +317,40 @@ function parse_system(exprs)
...
+and for the left-hand side, it returns either a empty vector `Any[]` or
```
```suggestion
and for the left-hand side, it returns either an empty vector `Any[]` or
```

src/macros.jl
```diff
@@ -317,9 +317,40 @@ function parse_system(exprs)
...
+`[(:E_user, :E)]` where the first argument of the tuple corresponds to the value
+and the second to the field parameter.
```
```suggestion
and the second argument of the tuple corresponds to the field parameter.
```

src/macros.jl
```diff
@@ -317,9 +317,40 @@ function parse_system(exprs)
...
+
+### Input
+
+- `equation` --
```
schillic: description missing

src/macros.jl
```diff
@@ -317,9 +317,40 @@ function parse_system(exprs)
...
+- `state`   -- state variable
+- `input`   -- input variable
+- `noise`   -- noise variable
+- `dim`     -- dimensionality
```
```suggestion
- `dim`     -- dimension
```
?

src/macros.jl
```diff
@@ -317,9 +317,40 @@ function parse_system(exprs)
...
+- `AT`      -- abstract system type
+
+### Output
+
+Two arrays of tuples containing the field and value parameter for the right-hand
```
```suggestion
Two arrays of tuples containing the value and field parameters for the right-hand
```

src/macros.jl
ueliwechsler: `dim` can also be `(1,2,3)`, i.e. multiple dimensions, so maybe `dimensions`. I used `dimensionality` because it was used in another docstring for the same variable https://github.com/JuliaReach/MathematicalSystems.jl/blob/e20619e2fa9d0d5e334a719facf37f8bd67d2362/src/macros.jl#L658

schillic: Okay, I put the "?" because I was not sure.

ueliwechsler: Then I will keep it for now and if need be, we can change it later on.

PR #170, opened 2020-02-16
add method for :(-) operator for IdentityMultiple with one argument

ueliwechsler:
Enables `-I(2)`.

### DIFF
--- a/src/identity.jl
+++ b/src/identity.jl
@@ -97,7 +97,7 @@ function Base.setindex!(𝐼::IdentityMultiple, X, inds...)
     error("cannot store a value in an `IdentityMultiple` because this type is immutable")
 end
 
-Base.:(-)(I::IdentityMultiple) = IdentityMultiple(-𝐼.M, 𝐼.n)
+Base.:(-)(𝐼::IdentityMultiple) = IdentityMultiple(-𝐼.M, 𝐼.n)
 Base.:(*)(x::Number, 𝐼::IdentityMultiple) = IdentityMultiple(x * 𝐼.M, 𝐼.n)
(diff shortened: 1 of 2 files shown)


src/identity.jl
```diff
@@ -97,6 +97,7 @@ function Base.setindex!(𝐼::IdentityMultiple, X, inds...)
     error("cannot store a value in an `IdentityMultiple` because this type is immutable")
 end
 
+Base.:(-)(I::IdentityMultiple) = -1*I
```
schillic: Hm, what about
```suggestion
Base.:(-)(im::IdentityMultiple) = IdentityMultiple(-im.M, im.n)
```

ueliwechsler: I thought it does not matter since it dispatches on `Base.:(*)(x::Number, 𝐼::IdentityMultiple) = IdentityMultiple(x * 𝐼.M, 𝐼.n)`. I will change it.

src/identity.jl
```diff
@@ -97,6 +97,7 @@ function Base.setindex!(𝐼::IdentityMultiple, X, inds...)
     error("cannot store a value in an `IdentityMultiple` because this type is immutable")
 end
 
+Base.:(-)(I::IdentityMultiple) = IdentityMultiple(-𝐼.M, 𝐼.n)
```
schillic: I != 𝐼 :stuck_out_tongue_winking_eye:

src/identity.jl
schillic: It should matter because a multiplication is more general than a negation.

src/identity.jl
ueliwechsler: thank you! I think I will go to bed now 🙈

mforets: This italic symbol in the code was probably a bad choice :smile:

PR #174, opened 2020-02-18
Add missing tests for scalar corner cases

### LINKED ISSUE #173: Add tests for scalar corner cases in @system
See https://github.com/JuliaReach/MathematicalSystems.jl/pull/168#issue-375862105.

Add tests after #168 is merged.

ueliwechsler:
closes #173 
and failing test work (it failed due to resolving #169)

### DIFF
--- a/test/@system.jl
+++ b/test/@system.jl
@@ -255,12 +255,12 @@ end
 # =======================
 @testset "@system with for an initial-value problem" begin
     # continuous ivp in floating-point
-    s = IVP(LinearContinuousSystem(I(-1.0, 1)), Interval(-1, 1))
-    @test @system(x' = -1.0x, x(0) ∈ Interval(-1, 1)) == s && eltype(s.A) == Float64
+    ivp = IVP(LinearContinuousSystem(I(-1.0, 1)), Interval(-1, 1))
+    @test @system(x' = -1.0x, x(0) ∈ Interval(-1, 1)) == ivp && eltype(ivp.s.A) == Float64
 
     # discrete ivp in floating-point
-    s = IVP(LinearDiscreteSystem(I(-1.0, 1)), [1])
-    @test @system(x⁺ = -x, x(0) ∈ [1]) == s && eltype(s.A) == Float64
+    ivp = IVP(LinearDiscreteSystem(I(-1.0, 1)), [1])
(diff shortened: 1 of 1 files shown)


test/@system.jl
```diff
@@ -252,7 +259,7 @@ end
     @test @system(x' = -1.0x, x(0) ∈ Interval(-1, 1)) == s
 
     # similar for integers
-    s = IVP(LinearDiscreteSystem(I(-1, 1)), [1])
+    s = IVP(LinearDiscreteSystem(I(-1., 1)), [1])
     @test @system(x⁺ = -x, x(0) ∈ [1]) == s
```
schillic: I think the integer was on purpose (see the comment).

ueliwechsler: Can you point me to the comment you are referring to?

schillic: the first line in the code excerpt :wink:

ueliwechsler: 🙈

ueliwechsler: Actually, I don't think we should have this distinction in the test for `IVPs` because this has nothing to do with the creation of a `IVP` but with the design decision from #139

ueliwechsler: https://github.com/JuliaReach/MathematicalSystems.jl/blob/d9e9331fb6add240931f92cd88e82948750fc8a1/src/macros.jl#L722-L724

schillic: Good catch. Then we should remove the comment and check that the type is indeed `Float64` I think.

ueliwechsler: Like this?

schillic: Not quite because previously the test also worked, probably because `==` converts automatically. Maybe something like (not tested)
```julia
    s = @system(x⁺ = -x, x(0) ∈ [1])
    @test s == IVP(LinearDiscreteSystem(I(-1.0, 1)), [1]) && eltype(s.A) == Float64
```

schillic: Note that you need to swap the definition of `s` for this test to work.

ueliwechsler: Sure, I tried to do a hasty quick fix. I am on it

schillic: And also `s` does not have a field `A` (my bad).

schillic: > Note that you need to swap the definition of s for this test to work.

I fixed that.

mforets:
Thanks!

PR #175, opened 2020-02-18
Fix scalar @system by converting arguments to matrices

ueliwechsler:
Thinking about other solutions to handle scalar system in `@system`.

The problems with the constructor for scalar systems could be fixed by converting the argument with `hcat` and `vcat` to a `matrix` respectively a `vector`.

Basically, the same we did in the constructor for  `Number` arguments.
Which means, that from the point of the `@system` macro, the scalar constructor would not be needed.

### DIFF
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -517,19 +517,19 @@ julia> using MathematicalSystems: extract_sum
 
 julia> extract_sum([:(A1*x)], :x, :u, :w)
 1-element Array{Tuple{Any,Symbol},1}:
- (:(hcat(:A1)), :A)
+ (:(hcat(A1)), :A)
 
 julia> extract_sum([:(A1*x), :(B1*u), :c], :x, :u, :w)
 3-element Array{Tuple{Any,Symbol},1}:
- (:(hcat(:A1)), :A)
- (:(hcat(:B1)), :B)
- (:(vcat(:c)), :c)
+ (:(hcat(A1)), :A)
+ (:(hcat(B1)), :B)
+ (:(vcat(c)), :c)
 
 julia> extract_sum([:(A1*x7), :( B1*u7), :( B2*w7)], :x7, :u7, :w7)
 3-element Array{Tuple{Any,Symbol},1}:
- (:(hcat(:A1)), :A)
- (:(hcat(:B1)), :B)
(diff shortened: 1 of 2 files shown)


mforets:
I like the idea in this PR :+1: 

My impression is that handling scalars in`@system` was more problematic than we thought.. I hope we can get to grips with that feature.. or throw it out of the window :laughing: 

NB: I just fixed the conflicts wrt `master` for the test script, so you may have to `git pull` on your local branch.

ueliwechsler:
> My impression is that handling scalars in`@system` was more problematic than we thought

Yes! We did not consider that as a result of having both scalar and matrices there occur corner cases. Therefore, it simplifies if we convert everything to an `Array` before calling the constructor.
AFAIK,  the only place where we don not convert the scalar is
https://github.com/JuliaReach/MathematicalSystems.jl/blob/d71a2841341f1776b6f8ab8d9fd0b2e9b0b91906/src/macros.jl#L412

ueliwechsler:
On  a different topic. Can I get some help with the doctest? 🙈 
`doctest(MathematicalSystems, fix=true)` does not work for me.

schillic:
> On a different topic. Can I get some help with the doctest? see_no_evil
> `doctest(MathematicalSystems, fix=true)` does not work for me.

Oh yes, I also noticed that at some point. I think the reason is that we do not always write `using MathematicalSystems` in the doctests.
In conclusion, I think you need to fix them manually (or add this command to the top of each doctest; it can be hidden with `# hide` I think).

Here is an example from the doctest errors:
```julia
│ extract_sum([:(A1*x)], :x, :u, :w)
│ 
│ Evaluated output:
│ 
│ 1-element Array{Tuple{Any,Symbol},1}:
│  (:(hcat(A1)), :A)
│ 
│ Expected output:
│ 
│ 1-element Array{Tuple{Any,Symbol},1}:
│  (:(hcat(:A1)), :A)
```

Here there is a colon (`:`) that apparently got removed.

Also notice that when @mforets merged the latest changes here, other tests started breaking.

ueliwechsler:
> Here there is a colon (`:`) that apparently got removed.
> 
> Also notice that when @mforets merged the latest changes here, other tests started breaking.

Thanks a lot. I did not notice it!

PR #176, opened 2020-02-18
Handle arbitrary order of  set / rhs term specifcation in @system

ueliwechsler:
Closes an element of #133 

The order of the set specification can be arbitrary now.

Additionally, it allows for arbitrary order of rhs term specification in the dynamic equation.

See test for an example.

### DIFF
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -675,11 +675,11 @@ to `order`.
 ### Example
 
 ```jldoctest
-julia> using MathematicalSystems: sort
+julia> using MathematicalSystems
 
 julia> parameters= [(:U1, :U), (:X1, :X), (:W1, :W)];
 
-julia> sort(parameters, (:X, :U, :W)
+julia> MathematicalSystems.sort(parameters, (:X, :U, :W))
 3-element Array{Tuple{Any,Symbol},1}:
  (:X1, :X)
  (:U1, :U)
@@ -687,7 +687,7 @@ julia> sort(parameters, (:X, :U, :W)
 
 julia>  parameters= [(:const, :c), (:A, :A)];
 
-julia> sort(parameters, (:A, :B, :c, :D)
+julia> MathematicalSystems.sort(parameters, (:A, :B, :c, :D))
 2-element Array{Tuple{Any,Symbol},1}:
  (:A, :A)
  (:const, :c)
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -664,7 +664,7 @@ end
 """
     sort(parameters::Vector, order::Tuple)
 
-Returns a sorted vector `parameters` according to `order`.
+Filter and sort the vector `parameters` according to `order`.
 `parameters` is a vector that contains tuples where the second element of each
 `Tuple` is considered for the sorting according to `order`.
 
@@ -674,11 +674,11 @@ If a value of `order` is not contained in `parameters` the corresponding entry o
 ### Input
 
 - `parameters` -- vector of tuples
-- `order`    -- tuple of symbols
+- `order`      -- tuple of symbols
 
 ### Output
 
-Sorted `parameters` vector according to `order`.
+A new vector corresponding to `parameters` filtered and sorted according to `order`.
 """
 function sort(parameters::Vector, order::Tuple)
     order_parameters = []
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -656,6 +656,42 @@ function extract_set_parameter(expr, state, input, noise) # input => to check se
     throw(ArgumentError("the set entry $(expr) does not have the correct form `x_ ∈ X_`"))
 end
 
+# Sort the parameter `parameters` such that it has the order `order` (e.g. (:X, :U, :W))
+# allowing for arbitrary order of  parameters which is a array of tuple, where the second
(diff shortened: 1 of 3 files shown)


src/macros.jl
```diff
@@ -656,6 +656,42 @@ function extract_set_parameter(expr, state, input, noise) # input => to check se
     throw(ArgumentError("the set entry $(expr) does not have the correct form `x_ ∈ X_`"))
 end
 
+# Sort the parameter `parameters` such that it has the order `order` (e.g. (:X, :U, :W))
+# allowing for arbitrary order of  parameters which is a array of tuple, where the second
+# element of each tuple is the relevant value for the ordering. If a value of `order`
+# is not contained in `parameters` it will be ignored.
```
schillic: I think you can remove this as it is covered by the docstring. If not, then merge it in there.

src/macros.jl
```diff
@@ -656,6 +656,42 @@ function extract_set_parameter(expr, state, input, noise) # input => to check se
...
+
+"""
+    sort(parameters::Vector, order::Tuple)
+
+Returns a sorted vector `parameters` according to `order`.
```
```suggestion
Filter and sort the vector `parameters` according to `order`.
```

src/macros.jl
```diff
@@ -656,6 +656,42 @@ function extract_set_parameter(expr, state, input, noise) # input => to check se
...
+`parameters` is a vector that contains tuples where the second element of each
+`Tuple` is considered for the sorting according to `order`.
+
+If a value of `order` is not contained in `parameters` the corresponding entry of
+`order` will be omitted.
+
+### Input
+
+- `parameters` -- vector of tuples
+- `order`    -- tuple of symbols
```
```suggestion
- `order`      -- tuple of symbols
```

src/macros.jl
```diff
@@ -656,6 +656,42 @@ function extract_set_parameter(expr, state, input, noise) # input => to check se
...
+
+### Output
+
+Sorted `parameters` vector according to `order`.
```
```suggestion
A new vector corresponding to `parameters` filtered and sorted according to `order`.
```

src/macros.jl
```diff
@@ -656,6 +656,37 @@ function extract_set_parameter(expr, state, input, noise) # input => to check se
     throw(ArgumentError("the set entry $(expr) does not have the correct form `x_ ∈ X_`"))
 end
 
+"""
+    sort(parameters::Vector, order::Tuple)
```
mforets: If possible I would prefer to see some types annotations,

```julia
    sort(parameters::Vector{Tuple{Any, Symbol}}, order::NTuple{N, Symbol}) where {N}
```

src/macros.jl
```diff
@@ -656,6 +656,37 @@ function extract_set_parameter(expr, state, input, noise) # input => to check se
...
+
+### Input
+
+- `parameters` -- vector of tuples
+- `order`      -- tuple of symbols
+
+### Output
+
+A new vector corresponding to `parameters` filtered and sorted according to `order`.
+"""
```
mforets: Can you please add a small `Example` section?

src/macros.jl
```diff
@@ -656,6 +656,63 @@ function extract_set_parameter(expr, state, input, noise) # input => to check se
...
+
+### Note
+
+`parameters` is a vector that contains tuples where the second element of each
+`Tuple` is considered for the sorting according to `order`.
+
+If a value of `order` is not contained in `parameters` the corresponding entry of
+`order` will be omitted.
+"""
+function sort(parameters::Vector{<:Tuple{Any, Symbol}}, order::NTuple{N, Symbol}) where {N}
```
schillic: You have to add `import Base: sort`. Otherwise you get namespace conflicts with `Base`. After that you should remove the prefix `MathematicalSystems.` in the doctests.

src/macros.jl
```diff
@@ -656,6 +656,63 @@ function extract_set_parameter(expr, state, input, noise) # input => to check se
...
+
+
+### Input
+
+- `parameters` -- vector of tuples
+- `order`      -- tuple of symbols
+
+### Output
+
+A new vector of tuplescorresponding to `parameters` filtered and sorted according
```
```suggestion
A new vector of tuples corresponding to `parameters` filtered and sorted according
```

src/macros.jl
```diff
@@ -656,6 +656,63 @@ function extract_set_parameter(expr, state, input, noise) # input => to check se
...
+
+julia> MathematicalSystems.sort(parameters, (:A, :B, :c, :D))
+2-element Array{Tuple{Any,Symbol},1}:
+ (:A, :A)
+ (:const, :c)
+```
+
+### Note
+
+`parameters` is a vector that contains tuples where the second element of each
```
```suggestion
`parameters` is a vector that contains tuples whose second element
```

src/macros.jl
```diff
@@ -656,6 +656,63 @@ function extract_set_parameter(expr, state, input, noise) # input => to check se
...
+`Tuple` is considered for the sorting according to `order`.
```
```suggestion
is considered for the sorting according to `order`.
```

src/macros.jl
```diff
@@ -656,6 +656,63 @@ function extract_set_parameter(expr, state, input, noise) # input => to check se
...
+
+If a value of `order` is not contained in `parameters` the corresponding entry of
```
```suggestion
If a value of `order` is not contained in `parameters`, the corresponding entry of
```

Issue #181, opened 2020-02-21
Error handling for repeated variable names (cont.)


mforets:
In https://github.com/JuliaReach/MathematicalSystems.jl/pull/164 we added default variable names in `parse_system` and an error message is sent in case of conflict. However, the noise variable is read a posteriori so there is still a possible conflict. As a consequence the following error is a bit cryptic and can be improved:

```julia
julia> @system(x⁺ = Ax + Bu + Du, input: u)
ERROR: ArgumentError: the entry (:A, :B, :B) does not match a `MathematicalSystems.jl` structure
```

ref: https://github.com/JuliaReach/MathematicalSystems.jl/pull/164#discussion_r378106601

An approach suggested [here](https://github.com/JuliaReach/MathematicalSystems.jl/pull/164#discussion_r380758870) is to make the error checks in `extract_sum`.

mforets:
> After thinking everything through more again more carefully, I think you are right.
> 
> In order for me to understand what is going on, I like to write everything I know down
> (sorry for the long comment)
> 
> As far as i know, this PR should give reasonable error messages for cases where
> a specified `input` or `noise` is added which does collide with another
> 
> ```julia
> A = B = D = rand(2,2)
> @system(a⁺= Aa + Bw, input:a)
> ERROR: ArgumentError: state and input variables have the same name `a`
> # instead of the old error message
> ERROR: ArgumentError: the entry (:A, :D) does not match a `MathematicalSystems.jl` structure
> @system(x⁺= Ax + Bx, input:x)
> ERROR: ArgumentError: state and input variables have the same name `x`
> @system(x⁺= Ax + Bu + Bw, input:u, noise:u)
> ERROR: ArgumentError: input and noise variables have the same name `u`
> ```
> 
> without throwing exceptions where the input is "overwritten" with the noise variable
> (and the noise not defined) or vice versa where the input is overwritten
> 
> ```julia
> # we can use 
> @system(x⁺= Ax + Bw, input:w) #creates a LinearControlDiscreteSystem system
> system(x⁺= Ax + Bu, noise:u) # 
> # instead of 
> @system(x⁺= Ax + Bw, input:w, noisestr:evry_but_w)
> ```
> 
> in the `@system(x⁺= Ax + Bw, input:w)`, the input and the noise variable are both `:w` but
> since we do not "use" the noise variable we accept that "inconsistency" for shortness of the
> macro.
> 
[...]

PR #186, opened 2020-02-22
Error handling for repeated variable names (cont.)

ueliwechsler:
More work on #181 

I added some more tests.

There are two tests that are commented out. I don't know how to test for these cases in a general way.
I suspect we will not be able to check them with the current setup (i.e. while allowing `@system(x' = Ax + Bw, input:w)` to work )


src/macros.jl
```diff
@@ -577,19 +577,27 @@ function extract_sum(summands, state::Symbol, input::Symbol, noise::Symbol)
     state_dim = 1
     got_state_dim = false
 
+    num_state_assignements = 0
+    num_input_assignements = 0
+    num_noise_assignements = 0
+    num_const_assignements = 0
```
schillic: typo: assignments

test/@system.jl
```diff
@@ -45,6 +45,18 @@ f1(x, u, w) = x'*x + u'*u + w'*w
         sys = @system(x' = c + Ax + Bu , u∈U, x∈X)
         @test sys == ConstrainedAffineControlContinuousSystem(A,B,c,X,U)
     end
+
+    @testset "@system for incorrect variable handling" begin
+        @test_throws ArgumentError("state and input variables have the same name `a`") @system(a⁺= Aa + Bw, input:a)
+        @test_throws ArgumentError("state and input variables have the same name `x`") @system(x⁺= Ax + Bx, input:x)
+        @test_throws ArgumentError("input and noise variables have the same name `u`") @system(x⁺= Ax + Bu + Bw, input:u, noise:u)
+        @test_throws ArgumentError("there is more than one input term `w`") @system(x⁺= Ax + Bw + Dw, input:w)
+        @test_throws ArgumentError("there is more than one input term `u`") @system(x⁺= Ax + Bu + Bu,  noise:u)
```
```suggestion
        @test_throws ArgumentError("there is more than one input term `u`") @system(x⁺= Ax + Bu + Bu, noise:u)
```

test/@system.jl
```diff
@@ -45,6 +45,18 @@ f1(x, u, w) = x'*x + u'*u + w'*w
...
+        @test_throws ArgumentError("there is more than one state term `x`") @system(x⁺= Ax + Bx)
+        @test_throws ArgumentError("there is more than one constant term")  @system(x⁺= Ax + c + c)
+        # @test_throws ArgumentError("input and noise variables have the same name `u`") @system(x⁺= Ax + Bu + Bw,  noise:u)
```
schillic: #### Summary
This command (`@system(x⁺= Ax + Bu + Bw, noise:u)`) throws
``ArgumentError: the entry (:A, :B, :c) does not match a `MathematicalSystems.jl` structure``
The problem is that `Bw` is parsed as a constant name in this case. Apparently we parse differently if an `input` name is defined explicitly, and also get confused when the same matrix occurs several times. Why is that?

The following work:
* `@system(x⁺= Ax + Bu + Bw)`  # no `, input: u`
* `@system(x⁺= Ax + Bu + B*w, input:u)`  # `*`
* `@system(x⁺= Ax + Bu + Dw, input: u)`  # `D` instead of `B`
* `@system(x⁺= Du + Aw + Bx, input: u)`  # different order and matrix names work

ueliwechsler: > Apparently we parse differently if an `input` name is defined explicitly

If the input is not defined, it does not throws an error because at the moment we only check for equality of variables if the are explicitly defined (see earlier discussions).
Additionally, we cannot check if `Bw` is a constant term or a noise term without the noise correctly defined due to the ambiguity of the entry. 
```julia
@system(x⁺= Ax + Bu + B*w, noise:u)
ERROR: ArgumentError: in the dynamic equation, the expression B * w does not contain the state x, the input u or the noise term u
Stacktrace:
 [1] top-level scope at none:0
```

> and also get confused when the same matrix occurs several times

This should and as far as I know does not happen.
I

schillic: Sorry, I do not remember this conversation, so I cannot comment here.

ueliwechsler: I was referring to the comments in this thread https://github.com/JuliaReach/MathematicalSystems.jl/pull/155#issuecomment-581594594
The design decision mentioned here https://github.com/JuliaReach/MathematicalSystems.jl/pull/155#issuecomment-584037527
makes the checks much harder than in the implementation presented in  #155.

To add to your summary, the case `@system(x⁺= Ax + Bu + Bw, noise:u)` is still a problem but using asterisks results in proper error messages for example `@system(x⁺= Ax + Bu + B*w, noise:u)`.

I propose merging this pull requests (once I've fixed the doctests) and opening another issue for the remaining error.

mforets:
Thanks a lot!

PR #191, opened 2020-02-23
#190 - Add getter function for system in IVP

### LINKED ISSUE #190: Add getter function for system in IVP

mforets:
Closes #190.


mforets:
> and what is the system of a system?

it can be defined as the identity for non-IVPs (not in this PR..)

PR #217, opened 2020-10-30
Return nothing for getter methods that are not supported

### LINKED ISSUE #200: Add better error messages for getter function which are not supported
It's a small thing but consider an unconstrained system
```julia
sys = @system(x' = Ax + Bu)
```
and try to access a nonexisting field, the error message could be improved.
```julia
stateset(a)
ERROR: MethodError: no method matching stateset(::LinearControlDiscreteSystem{Float64,Array{Float64,2},Array{Float64,2}})
Closest candidates are:
  stateset(::ConstrainedContinuousIdentitySystem) at C:\Users\wueli\.julia\dev\MathematicalSystems\src\systems.jl:71
  stateset(::ConstrainedDiscreteIdentitySystem) at C:\Users\wueli\.julia\dev\MathematicalSystems\src\systems.jl:71
  stateset(::ConstrainedLinearContinuousSystem) at C:\Users\wueli\.julia\dev\MathematicalSystems\src\systems.jl:327
  ...
Stacktrace:
 [1] top-level scope at none:0
```

For example we could return:
```julia
ERROR: the state set cannot be accessed for a system of  type `LinearControlDiscreteSystem` 
```

ueliwechsler:
closes #200 

Following the proposal of @schillic, I implemented a default case for all getters that returns `nothing` if the method is not implemented. Additionally, I tried to give the unit tests for discrete and continuous systems more structure (which resulted in a large PR...sorry 😁)

As a by-product, I also added the supertype `AbstractSystem` to the `SystemWithOutput` type.

### DIFF
--- a/test/continuous.jl
+++ b/test/continuous.jl
@@ -1,248 +1,301 @@
+# linear systems
+E = [0. 1; 1 0]
+A = [1. 1; 1 -1]
+B = Matrix([0.5 1.5]')
+C = [1.; 1.]
+D = [1. 2; 0 1]
+X = Line([1., -1], 0.) # line x = y
+U = Interval(0.9, 1.1)
+W = BallInf(zeros(2), 2.0)
+sd = 2
+id = 1
+nd = 2
+
+# polynomial system
+@polyvar x y
+p = 2x^2 - 3x + y
+sdp = 2
+
+# blackbox system
+add_one(x) = x .+ 1
+add_one(x, u) = x .+ 1 .+ u
+add_one(x, u, w) = x .+ 1 .+ u .+ w
+state = [1.0; 2.0]
+input = 1
+noise = [3.0; 1.0]
+statePlusOne = add_one(state)
+stateInputPlusOne = add_one(state, input)
+stateInputNoisePlusOne = add_one(state, input, noise)
+
+# Scalar System
+a = 1.; b = 2.; c = 0.1; d = 3.; Xs = 1; Us = 2; Ws = 3; e = 2.;
+As = [a][:,:]; Bs = [b][:,:]; Cs = [c]; Ds = [d][:,:]; Es = [e][:,:]
+
+
+
 @testset "Continuous identity system" begin
-    for sd in 1:3
-        s = ContinuousIdentitySystem(sd)
-        @test statedim(s) == sd
-        @test inputdim(s) == 0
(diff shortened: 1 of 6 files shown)


mforets:
> Would nothing work as a default "Universe" type with respect to the ReachabilityAnalysis API?

Yes, we can handle that easily by making a new dispatch on structs with a `Nothing` field, which should behave as if it was a `LazySets.Universe`.

In fact after parsing the user input, (linear) reachability algorithms normalize systems to only `x' = Ax` (LCS) or the canonical form `x' = Ax + u` (CLCCS) with state and input constraints on `x` and `u` respectively. This covers all relevant cases. The normalization step is currently implemented [here](https://github.com/JuliaReach/ReachabilityAnalysis.jl/blob/master/src/Continuous/normalization.jl#L255). But this normalization would also be useful for another package. I think a good idea is to port such normalization to MathematicalSystems.jl.

blegat:
There could be a way to get an error too
```julia
function well_chosen_function_name(func::Function, system, args...)
    ret = func(system, args...)
    if ret === nothing
        error("$func is not implemented by $(typeof(system)).")
    end
    return ret
end
```

test/continuous.jl
```diff
@@ -1,248 +1,301 @@
...
+stateInputPlusOne = add_one(state, input)
+stateInputNoisePlusOne = add_one(state, input, noise)
+
+# Scalar System
+a = 1.; b = 2.; c = 0.1; d = 3.; Xs = 1; Us = 2; Ws = 3; e = 2.;
+As = [a][:,:]; Bs = [b][:,:]; Cs = [c]; Ds = [d][:,:]; Es = [e][:,:]
+
+
+
 @testset "Continuous identity system" begin
```
schillic, line 36: I learned something new here ([doc](https://docs.julialang.org/en/v1/stdlib/Test/#Test.@testset)): You can nest `@testset`s! In case of no error, it will just print the outer test sets. In case of an error, it will print the inner test sets as well. I just wanted to share this.

schillic (approved):

test/continuous.jl
ueliwechsler: Thanks for the review!
Yes, the nesting is nice, it makes the testing output so much neater.

mforets (approved):

mforets:
Merging..

PR #220, opened 2020-11-28
Plot recipe for VectorField

schillic:
This is open for discussion.

Usage:
```julia
julia> f(x) = [2*x[2], -x[1]]
julia> V = MathematicalSystems.VectorField(f)
julia> plot(V)  # uses a default grid of 20x20 points for the range [-3, 3] x [-3, 3]
```

```julia
julia> plot(V, x=range(-3, stop=3, length=11), y=range(-3, stop=3, length=11))  # pass a custom grid
```

EDIT: For reference, this is the result of the default quiver plot (note that here the function `f` needs to take two arguments and in my approach it needs to take a vector):

```julia
f(x, y) = [2*x, -y]
x = range(-3, stop=3, length=21)
y = range(-3, stop=3, length=21)
x_full = [xi for xi in x for yj in y]
y_full = [yj for xi in x for yj in y]
quiver(x_full, y_full, quiver=f)
```

### DIFF
--- a/src/vector_field.jl
+++ b/src/vector_field.jl
@@ -163,3 +163,71 @@ function VectorField(sys::AbstractContinuousSystem)
     end
     return VectorField(field)
 end
+
+function _position_value_list(V::VectorField, x, y)
+    x1 = x[1]
+    y1 = y[1]
+    N_POS = typeof(x1)
+    N_VAL = typeof(V([x1, y1])[1])
+    m = length(x) * length(y)
+
+    sx = Vector{N_POS}(undef, m)
+    sy = similar(sx)
+    tx = Vector{N_VAL}(undef, m)
+    ty = similar(tx)
+    k = 1
+    max_entry = N_VAL(-Inf)
+    for xi in x
+        for yj in y
+            sx[k] = xi
+            sy[k] = yj
+            tx_k, ty_k = V([xi, yj])
+            max_entry = max(max_entry, abs(tx_k), abs(ty_k))
+            tx[k] = tx_k
+            ty[k] = ty_k
+            k += 1
+        end
+    end
+
+    # normalize
+    if max_entry > zero(N_VAL)
+        step_x = (maximum(x) - minimum(x)) / length(x)
+        step_y = (maximum(y) - minimum(y)) / length(y)
+        max_entry_x = max_entry / step_x
+        max_entry_y = max_entry / step_y
+        for k in 1:m
+            tx[k] /= max_entry_x
+            ty[k] /= max_entry_y
+        end
+    end
+
+    # filter out (0, 0) entries
+    for k in m:-1:1
+        if tx[k] == ty[k] == zero(N_VAL)
+            deleteat!(sx, k)
+            deleteat!(sy, k)
+            deleteat!(tx, k)
+            deleteat!(ty, k)
+        end
+    end
+
+    # center arrows in the grid points
+    for k in 1:length(sx)
+        sx[k] -= tx[k] / 2
+        sy[k] -= ty[k] / 2
+    end
+
+    return sx, sy, tx, ty
+end
+
+### plot recipe
+@recipe function plot(V::VectorField;
+                      x=range(-3, stop=3, length=21),
+                      y=range(-3, stop=3, length=21))
+    seriestype := :quiver
+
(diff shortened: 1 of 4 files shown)


src/vector_field.jl
```diff
@@ -163,3 +163,71 @@ function VectorField(sys::AbstractContinuousSystem)
...
+    end
+
+    return sx, sy, tx, ty
+end
+
+### plot recipe
+@recipe function plot(V::VectorField;
+                      x=range(-3, stop=3, length=21),
+                      y=range(-3, stop=3, length=21))
+    seriestype := :quiver
```
mforets, line 227: check the the vector field is 2D and give an error otherwise?

src/vector_field.jl
```diff
@@ -163,3 +163,71 @@ function VectorField(sys::AbstractContinuousSystem)
...
+        sx[k] -= tx[k] / 2
+        sy[k] -= ty[k] / 2
+    end
+
+    return sx, sy, tx, ty
+end
+
+### plot recipe
+@recipe function plot(V::VectorField;
+                      x=range(-3, stop=3, length=21),
```
mforets, line 225: i think it can be useful if we also allow the kwargs `xmin`, `xmax`, `ymin`, `ymax`? or can we just take `xlims` and `ylims` which are already valid Plots kwargs?

src/vector_field.jl
```diff
@@ -163,3 +163,71 @@ function VectorField(sys::AbstractContinuousSystem)
...
 end
+
+function _position_value_list(V::VectorField, x, y)
+    x1 = x[1]
+    y1 = y[1]
+    N_POS = typeof(x1)
+    N_VAL = typeof(V([x1, y1])[1])
+    m = length(x) * length(y)
+
+    sx = Vector{N_POS}(undef, m)
```
mforets, line 174: does the `s` or the `t` represents widths along each direction of the arrows? can you please add a short description of what the function is doing?

mforets (commented):
Nice!!

src/vector_field.jl
schillic: So define a default number of grid points you mean (the `length` argument in my `range` above)? That is probably reasonable because when I experimented with different numbers, there is only a limited reasonable range of 10-30 grid points in each dimension.

src/vector_field.jl
schillic: I think when I wrote the code I had `s` and `t` for source and target coordinates of the arrow in mind, but I may be wrong.

schillic:
I will probably not have time to work on this in the near future. Feel free to modify.

schillic:
Side note: `VectorField` is not exported. I do not remember whether that was on purpose or an oversight.

mforets:
> I will probably not have time to work on this in the near future. Feel free to modify.

:+1:  me too and since the others replied with thumbs up and the PR is functional, I suggest to proceed merging it and addressing a documentation PR separately. (Please note that the build is unrelated and only affects Julia v1.0).

PS: The new function may interest @dfcaporale

mforets (approved):

src/vector_field.jl
schillic: Actually, you can also plot a 2D projection. For that you could be able to pass the two dimensions that you want to plot, probably defaulting to `1` and `2`.

schillic: I did that now. You can do `f(x) = [2*x[2], -x[1], 7]` in my example above and `plot(V)` will give the same result. You can do `plot(V, dims=[1, 3])` to change the dimensions.

src/vector_field.jl
schillic: I combined and renamed the arguments to `grid_points`, which should be more understandable. (`grid` is already reserved by `Plots`.)

src/vector_field.jl
schillic: No, I was wrong. I renamed the variables and added a small documentation (just as a comment).

schillic:
I fixed the build issue and addressed your comments.
I do not mind keeping this open a bit until we agree on the API. I mainly wanted to share the code and not forget about it.

mforets:
Merging.. I don't mind to follow up with API changes / docs updates in subsequent PRs.

PR #222, opened 2020-12-04
add AffineControlContinuousSystem and AffineControlDiscreteSystem system

ueliwechsler:
Another missing type. I did create it a month ago, but should still worth a PR.

In retrospective, it could have been possibly nice to define every type as a constrained type and let the macro figure out if the user wants a constraint set or not.


test/@system.jl
```diff
@@ -209,7 +209,7 @@ end
 @testset "@system for affine discrete systems" begin
     @test @system(x⁺ = A*x + c) == AffineDiscreteSystem(A, c)
     @test @system(x⁺ = A1*x + c1) == AffineDiscreteSystem(A1, c1)
-    @test_throws ArgumentError @system(x⁺ = Ax + Bu + c) # not a system type
+    # @test_throws ArgumentError @system(x⁺ = Ax + Bu + c) # not a system type
```
schillic, line 212:
```suggestion
```

test/continuous.jl
```diff
@@ -141,6 +141,25 @@ end
...
+    s = AffineControlContinuousSystem(A, B, c)
+    @test statedim(s) == 2
+    @test inputdim(s) == 1
+    @test noisedim(s) == 0
+    for s = [s, typeof(s)]
+        @test !islinear(s) && isaffine(s) && !ispolynomial(s) && !isblackbox(s)
+        @test !isnoisy(s) && iscontrolled(s) && !isconstrained(s)
+    end
+    # Scalar System
+    a = 1.; b = 2.; c = 0.1; X = 1; U = 2
```
schillic, line 157:
```suggestion
    a = 1.; b = 2.; c = 0.1
```

schillic (approved):
I think this type of system is rather redundant because it can be modeled with a `LinearControlXSystem` where you incorporate `c` into `Bu`. Still, it is nice to parse it correctly.

mforets:
True, but it has some advantages too:

1) On the modeling side, it is not uncommon to split the input a constant and a time-varying case so having this directly as a template system structure is quite nice (a concrete example is small signal analysis in electronics https://en.wikipedia.org/wiki/Small-signal_model )

2) On the conservative discretization side we can reason, or dispatch, on `c` known to be a constant and `Bu` being a possibly uncertain input.

test/@system.jl
```diff
@@ -142,7 +142,7 @@ end
     @test @system(x' = A*x  + c) == AffineContinuousSystem(A, c)
     sys =  @system(z_1' = A*z_1 + B*v_1 + c1, z_1 ∈ X, v_1 ∈ U1, input:v_1)
     @test sys == ConstrainedAffineControlContinuousSystem(A, B, c1, X, U1)
-    @test_throws ArgumentError @system(x' = Ax + Bu + c) # not a system type
+    # @test_throws ArgumentError @system(x' = Ax + Bu + c) # not a system type
```
mforets, line 145: just checked w/master and the test works now :100: so we can uncomment it

```suggestion
   sys =  @system(x' = Ax + Bu + c)
    @test sys == AffineControlContinuousSystem(A, B, c)
```

mforets (approved):

mforets:
Merging.. then I'll tag a new release.

PR #223, opened 2020-12-04
Fix affine system corner case

### LINKED ISSUE #221: IVP macro should return affine system
```julia
julia> b = [1.0]
1-element Array{Float64,1}:
 1.0

julia> @system(x' = x + b)
LinearControlContinuousSystem{Float64,IdentityMultiple{Float64},IdentityMultiple{Float64}}([1.0], [1.0])
```
The answer is OK but I would expect an `AffineContinuousSystem` instead. This way we can use the type information to know that the affine term is a singleton (and not `Bu(t)` as in the LCCS which is much more general).

ueliwechsler:
closes #221 

Adding the condition that the stripped system is only parsed as a black-box system, if `f` is not an addition, multiplication or subtraction.

### DIFF
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -280,14 +280,20 @@ function _parse_system(exprs::NTuple{N, Expr}) where {N}
                 dynamic_equation = stripped
                 state_var = subject
                 AT = abstract_system_type
-                # if the system has the structure x_ = A_*x_ + B_*u_ ,
-                # handle u_ as input variable
+                # if the stripped system has the structure x_ = A_*x_ + B_*u_ or
+                # one of the other pattern, handle u_ as input variable
                 if @capture(stripped, (x_ = A_*x_ + B_*u_) |
                                       (x_ = x_ + B_*u_) |
                                       (x_ = A_*x_ + B_*u_ + c_) |
                                       (x_ = x_ + B_*u_ + c_) |
                                       (x_ = f_(x_, u_)) )
-                    input_var = u
+                    if (f == :+) || (f == :-) || (f == :*)
+                        # pattern x_ = f_(x_, u_) also catches the cases:
+                        # x_ = x_ + u_, x_ = x_ - u_ and x_ = x_*u_
+                        # where u_ doesn't necessarily needs to be the input
+                    else
+                        input_var = u
+                    end
                 end
 
             elseif @capture(ex, (dim = (f_dims_)) | (dims = (f_dims_)))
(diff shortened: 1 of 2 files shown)


src/macros.jl
```diff
@@ -280,14 +280,20 @@ function _parse_system(exprs::NTuple{N, Expr}) where {N}
...
                 if @capture(stripped, (x_ = A_*x_ + B_*u_) |
                                       (x_ = x_ + B_*u_) |
                                       (x_ = A_*x_ + B_*u_ + c_) |
                                       (x_ = x_ + B_*u_ + c_) |
                                       (x_ = f_(x_, u_)) )
-                    input_var = u
+                    if (f == :+) || (f == :-) || (f == :*)
+                        # pattern x_ = f_(x_, u_) also catches the cases:
+                        # x_ = x_ + u_, x_ = x_ - u_ and x_ = x_*u_
+                        # where u_ doesn't necessarily needs to be the input
```
schillic, line 293:
```suggestion
                        # where u_ doesn't necessarily need to be the input
```

src/macros.jl
```diff
@@ -280,14 +280,20 @@ function _parse_system(exprs::NTuple{N, Expr}) where {N}
                 dynamic_equation = stripped
                 state_var = subject
                 AT = abstract_system_type
-                # if the system has the structure x_ = A_*x_ + B_*u_ ,
-                # handle u_ as input variable
+                # if the stripped system has the structure x_ = A_*x_ + B_*u_ or
+                # one of the other pattern, handle u_ as input variable
```
schillic, line 284:
```suggestion
                # one of the other patterns, handle u_ as input variable
```

schillic (approved):
I think this is again only some heuristics.
The following does not work both in `master` and here.
```julia
julia> @system x' = 2 * f(x)
ERROR: ArgumentError: if there is only one term on the right-hand side, the state needs to be contained
```

ueliwechsler:
> I think this is again only some heuristics.

Yes and no. You could call `@system` itself a heuristic since we try to catch as many cases as possible. For your example, we would need some kind of recursion in the parsing of the system which we don't have implement.
But this PR is a fix for a case that returned the false system, it corrects an error in the implementation.

It might be worth to gather information on how the `@system` macro was used in the last year and from this information, discuss how the macro could be refactored. I suspect, there are some features that are not needed and leaving them out could simplify the implementation or enable to implement a structure as discussed in https://github.com/JuliaReach/MathematicalSystems.jl/issues/221.

mforets (approved):

test/@system.jl
```diff
@@ -139,7 +139,9 @@ end
 end
 
 @testset "@system for affine continuous systems" begin
-    @test @system(x' = A*x  + c) == AffineContinuousSystem(A, c)
+    b = [1.0]
+    @test @system(x' = x + b) == AffineContinuousSystem(I(1), b)
```
mforets, line 143: just checked that the vector form doesn't work (i'll open an issue):

```julia
julia> b = rand(10); @system(x' = x + b)
ERROR: AssertionError: checksquare(A) == length(c)
Stacktrace:
 [1] AffineContinuousSystem(::IdentityMultiple{Float64}, ::Array{Float64,1}) at /home/mforets/.julia/dev/MathematicalSystems/src/systems.jl:189
 [2] top-level scope at REPL[4]:1
```

mforets: please note that in pple the transformation can be done (without manually specifying `dim=10`), because the code would generate `I(n, n)` where`n=length(b)`.

ueliwechsler: I took me so long to figure out, what you meant by pple 😅
Yes, but from my experience it is a bit akward to get the length from the input vector and return it again in a macro.

PR #226, opened 2020-12-10
Add systems section

mforets:


mforets:
This is ready for review. I added a summary table for the systems types and some introductory text on how to use `@system` and `@ivp`.

docs/src/man/systems.md
```diff
@@ -0,0 +1,127 @@
...
+|$Ex' = Ax$|no|no|LACS|
+|$Ex' = Ax, x ∈ X$|yes|no|CLACS|
+|$x' = p(x)$|no|no|PCS|
+|$x' = p(x), x ∈ X$|yes|no|CPCS|
+|$x' = f(x)$|no|no|BBCS|
+|$x' = f(x), x ∈ X$|yes|no|CBBCS|
+|$x' = f(x, u)$|no|no|BBCCS|
+|$x' = f(x, u), x ∈ X, u ∈ U$|yes|yes|CBBCCS|
+|$x' = Ax + Dw$|no|no|NLCS|
+|$x' = Ax + Dw, x ∈ X, w ∈ W$|yes|yes|NCLCS |
```
schillic, line 117: Should there not be another column for noise? I thought `w ∈ W` is *not* an input constraint.

schillic (commented):
Is there a reason for the two tables? Are the names too long?

docs/src/man/systems.md
```diff
@@ -0,0 +1,127 @@
+# System types overview
+
+A convenient way to create a new system is with the `@system` macro. For example,
+the ODE $x'(t) = -2x(t)$ is simply `x' = -2x`, where $x'(t) := dx/dt$ is the derivative
```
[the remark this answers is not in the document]
mforets: yes

docs/src/man/systems.md
mforets: i kept just one column to simplify. there was no ambiguity because, if i'm not mistaken, the systems (currently, at least) constrain the input set iff they constrain the noise set.

mforets:
> Is there a reason for the two tables? Are the names too long?

yes. if i keep the long names they don't fit into a single row. otoh it would be nice that it's possible to click on the abbreviated names. i can add the hyperlinks later.

docs/src/man/systems.md
schillic: @ueliwechsler can you remember?

mforets: i added [a sentence](https://github.com/JuliaReach/MathematicalSystems.jl/pull/226/files#diff-c92c808924e08a9c9b4253a9b9cc6196a725c435ab6c9b5fe844c8605ffdefe0R95) to clarify this point

docs/src/man/systems.md
```diff
@@ -0,0 +1,128 @@
...
+@system(x' = -2x)
+```
+
+We can also add state constraints, say $x(t) ≥ 0.5$,
+
+```@example system_examples
+using LazySets
+
+H = HalfSpace([-1.0], -0.5) # x >= 0.5
+@system(x' = -2x, x ∈ H)
```
mforets, line 19: At some later point it would be nice that `@system(x' = -2x, x >= 0.5)` does the same thing.

ueliwechsler:
@schillic Sorry for not answering, I posted a response from my phone but did not see I had to submit it as well. 
> The clarification sentence sounds good to me.
A system is either constrained or not and depending on the "input type" (state, input and or noise) it has then state, input and or noise constraints.

docs/src/man/systems.md
schillic: > the systems (currently, at least) constrain the input set iff they constrain the noise set.

But does the same not apply to state constraints as well? With this argument, the second column is also redundant. When I read this table, I would get confused about what is the difference between
```
|$x' = Ax + Bu, x ∈ X, u ∈ U$|yes|yes|CLCCS|
|$x' = Ax + Dw, x ∈ X, w ∈ W$|yes|yes|NCLCS |
```
because syntactically they look equivalent. You have to know that we interpret `W` as noise and `U` as inputs, but if you are not aware of these conventions, you have no way of understanding that from the table.

PR #249, opened 2022-10-04
Use latest Julia version (v1.*) in CI

schillic:


blegat (commented):
Let's do the same as https://github.com/JuliaReach/MathematicalSets.jl/pull/24/ ?

PR #251, opened 2022-11-23
Rename algebraic systems into descriptor systems

blegat:
I believe the correct terminology is "algebraic" to talk about the **equation** but "descriptor" to talk about the **system**.


schillic (approved):
I have no idea about the terminology, but I am fine with changing it. This would be a breaking change I guess.

blegat:
Yes this is breaking unfortunately so that will need v0.13. We can wait a few days to see if there is any objection, there is no rush

mforets (approved):

schillic:
@blegat Should we merge this and make a new release?

blegat:
Yes, let's do it

PR #260, opened 2023-06-13
General "black-box" types for output maps

adrienbanse:
This PR follows this discussion: https://github.com/dionysos-dev/Dionysos.jl/issues/236. We feel like a general type for output maps is missing right now in MathematicalSystems.jl. 

Therefore, this PR implements four new types: 
- `BlackBoxMap`, a general type for output maps
- `ConstrainedBlackBoxMap`, a general type for output maps in which the state-space is constrained
- `BlackBoxControlMap`, a general type for control output maps
- `ConstrainedBlackBoxControlMap`, a general type for control output maps in which the state-space and input-space are constrained

These should be covered by the tests written in the PR, and the doc should be clean as well. 

@blegat

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -4,25 +4,17 @@ import LazySets
 DocMeta.setdocmeta!(MathematicalSystems, :DocTestSetup,
                     :(using MathematicalSystems); recursive=true)
 
-makedocs(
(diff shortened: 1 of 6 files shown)


schillic (approved):
Thanks!

PR #263, opened 2023-06-23
Update package bounds

schillic:


blegat (approved):

schillic:
I guess we should also make a patch release to publish this change. I will add it here.

PR #290, opened 2024-03-02
Fix ambiguities with LinearAlgebra

### LINKED ISSUE #283: Fix ambiguities
See the failing Aqua test.

schillic:
Closes #283.

I simply reused the generic implementation and did not think about whether the specializations could be more efficient.


blegat (approved):
This is tricky to maintain indeed, we have the same issue in MutableArithmetics

PR #312, opened 2024-11-08
Change :exact to ExactDiscretization in discretize

blegat:


schillic:
Good catch, but the interface is actually completely different. `algorithm` is the first argument:

https://github.com/JuliaReach/MathematicalSystems.jl/blob/11ee5eced6521155b07eb567b0d9f6dbec26a773/src/discretize.jl#L216-L217

schillic (approved):

PR #319, opened 2025-01-11
#209 - Define `state_matrix` and `affine_term` for maps

### LINKED ISSUE #209: Add getter functions for map fields
Add `state_matrix` (or just `matrix`) and `affine_term` (or `vector`) getter functions for concrete subtypes of `AbstractMap`.

schillic:
Closes #209.

### DIFF
--- a/src/maps.jl
+++ b/src/maps.jl
@@ -17,6 +17,7 @@ end
 statedim(m::IdentityMap) = m.dim
 outputdim(m::IdentityMap) = m.dim
 inputdim(::IdentityMap) = 0
+state_matrix(m::IdentityMap) = Id(m.dim)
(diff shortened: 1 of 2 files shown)


blegat:
I would find `constant_term` more natural than `affine_term`

schillic:
Okay, I followed the description in #209 and chose `state_matrix` and `affine_term` because we use these names for systems too. I am worried that using a different name for maps is rather confusing.

https://github.com/JuliaReach/MathematicalSystems.jl/blob/c1b8b0657f8a7beb0e329acc64a62df865f02261/src/systems.jl#L197-L198

blegat:
If we're already using it for systems then let's go for it

blegat (approved):

PR #368, opened 2026-03-31
Revamp macros definitions

mforets:


src/macros.jl
```diff
@@ -57,18 +57,23 @@ macro map(ex, args...)
...
-        else
-            return Expr(:call, IdentityMap, esc(:($(dimension))))
+        if length(args) > 1
+            X = @match args[2] begin
+                :($_ ∈ $X) => X
+                _ => nothing
+            end
+            if X !== nothing
+                return Expr(:call, ConstrainedIdentityMap, esc(:($(dimension))), esc(:($(X))))
+            end
```
schillic, line 74: This is different from the old code, which required that the constrained variable is called `x`. I think some cases are better now. Other cases should have been throwing an error. But this would require a more fundamental change. So I am fine with that new behavior.

```julia
julia> X = 5;

# master: ignores `y` as variable in constraint
julia> @map(y -> y, dim=2, y ∈ X)
IdentityMap(2)

julia> @map(x -> x, dim=2, y ∈ X)
IdentityMap(2)


# this branch
julia> @map(y -> y, dim=2, y ∈ X)  # accepts `y` as variable
ConstrainedIdentityMap{Int64}(2, 5)

julia> @map(x -> x, dim=2, y ∈ X)  # misses non-matching variable name
ConstrainedIdentityMap{Int64}(2, 5)
```

src/macros.jl
```diff
@@ -160,6 +160,11 @@ function _corresponding_type(AT::Type{<:AbstractSystem}, fields::Tuple)
     return TYPES[idx][1]
 end
 
+# Normalize Expr(:kw, ...) to Expr(:(=), ...) so that MLStyle quote patterns
+# like `:($lhs = $rhs)` match both `dim = 3` (parenthesized macro call, parsed
+# as :kw) and `dim = 3` (unparenthesized, parsed as :(=)).
```
schillic: The two occurrences of `dim = 3` are identical. Did you mean something else?

src/macros.jl
```diff
@@ -271,70 +291,75 @@ function _parse_system(exprs::NTuple{N,Expr}) where {N}
...
-            # TODO? handle equality, || @capture(ex, x_(0) = X0_)
-            if x != state_var
-                throw(ArgumentError("the initial state assignment, $x(0), does " *
-                                    "not correspond to the state variable $state_var"))
-            end
-            initial_state = X0
+        else
+            @match ex begin
+                :($x(0) ∈ $X0) => begin  # COV_EXCL_LINE
+                    # TODO? handle equality
```
```suggestion
                    # TODO handle equality (x(0) = x0)?
```

src/macros.jl
```diff
@@ -994,7 +1051,10 @@ macro ivp(expr...)
         # check if the @ivp macro is called with an AbstractSystem argument
         if parses_as_system(expr[1])
             sys = expr[1]
-            @capture(expr[2], x_(0) ∈ x0_) || throw(ArgumentError("malformed epxpression")) # TODO handle equality
+            x0 = @match expr[2] begin
+                :($x(0) ∈ $X0) => X0
+                _ => throw(ArgumentError("malformed epxpression")) # TODO handle equality
```
```suggestion
                _ => throw(ArgumentError("malformed epxpression")) # TODO handle equality (x(0) = x0)?
```

schillic (commented):
Amazing! I know this is marked as a draft, but I already had a look. Currently, this PR is a perfect refactoring (with one exception, see my comment below).

src/macros.jl
mforets: Agreed — I'll leave this as is. Thanks for double-checking the edge cases.

src/macros.jl
mforets: Good catch, that was a confusing example — both snippets rendered as the same text `dim = 3` even though they come from different parse paths. Clarified in 29c9d54 to show the actual macro-call forms (`@macro(dim = 3)` vs `@macro dim = 3`).

src/macros.jl
mforets: Applied, thanks.

src/macros.jl
mforets: Applied, thanks.

src/macros.jl
```diff
@@ -994,7 +1052,10 @@ macro ivp(expr...)
         # check if the @ivp macro is called with an AbstractSystem argument
         if parses_as_system(expr[1])
             sys = expr[1]
-            @capture(expr[2], x_(0) ∈ x0_) || throw(ArgumentError("malformed epxpression")) # TODO handle equality
+            x0 = @match expr[2] begin
+                :($x(0) ∈ $X0) => X0
+                _ => throw(ArgumentError("malformed epxpression")) # TODO handle equality (x(0) = x0)?
```
```suggestion
                _ => throw(ArgumentError("malformed expression")) # TODO handle equality (x(0) = x0)?
```

schillic (approved):

PR #374, opened 2026-06-06
Use latest version of JuliaFormatter again

schillic:
The latest JuliaFormatter behavior is sufficiently similar to v1. See #322.


schillic:
The build error is only with the Codecov upload. In other repos it used to work a few hours ago. Maybe just a temporary hiccup...

mforets (approved):

